PR #1, opened 2017-08-10
Publish

bratseth:
@kraune please review


kkraune:
this should be in gh-pages branch ?

bratseth:
People found that confusing, and we don't really need it now.
I suggest we set up something that moves master to that branch on every checkin.
Perhaps you could add master to gh-pages now to make the doc work today? I have to leave for the evening.

kkraune:
i will look into it today :-)

PR #7, opened 2017-08-12
Make the quick start as simple as possible.

aressem:
@kkraune 

FYI, @bratseth


kkraune:
nice! will try it out!

PR #13, opened 2017-08-16
Update streaming doc after recent changes

jobergum:
Remove allowvisitcache as default has been flipped, cache is controlled as with indexed search. 

Please review the last parts here about searchable-copies, my understanding is that the default searchable copies is equal to redundancy and better to simply remove searchable-copies entirely from the example.


kkraune:
doc on searchable copes says _Default value is 2, or redundancy, if higher_

kkraune:
I agree the searchable copies thing is confusing - and we can probably delete it as it defaults to redundancy. I seem to remember we added this for a reason, with bucket (de)-activation problems that @bjormel hit

PR #16, opened 2017-08-16
A few improvements to blog tutorial

jobergum:
Need to re-factor again after sample apps moved to a own repo, also we need to discuss the size of the sample data (1M docs).


jobergum:
Please merge @kkraune

kkraune:
👍

PR #34, opened 2017-08-24
Made some minor changes to typography and language

jonmv:
@kkraune please review and merge :]

### DIFF
--- a/documentation/vespa-quick-start.html
+++ b/documentation/vespa-quick-start.html
@@ -100,26 +100,25 @@ $ docker exec -it vespa bash
 
 <ul>
 <li>
-Try the <a href="tutorials/blog-search.html">blog search tutorial</a>
-to learn more about the different Vespa features.
+    Try the <a href="tutorials/blog-search.html">blog search tutorial</a>
+    to learn more about the different Vespa features.
 </li>
 
 <li>
-Developing with Vespa includes writing Java plugins for document and query processors;
-try <a href="jdisc/developing-applications.html">developing applications</a>
-to compile and test Vespa JDisc Container plugins.
+    Developing with Vespa includes writing Java plugins for document and query processors;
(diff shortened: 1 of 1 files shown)


jonmv:
Hmm ... the <code> elements are too red?

jonmv:
Hang on, need more linespacing between items.

documentation/vespa-quick-start.html
```diff
@@ -4,92 +4,121 @@
 ---
 
 <p>
-This article is tested using Docker for Mac - adapt as fit for other platforms.
-Also try the <a href="vespa-quick-start-centos.html">CentOS quick-start</a>.
-</p>
+This article is tested using Docker for Mac &mdash; adapt as you see fit for other platforms.
```
kkraune: why &amp;mdash; ?    makes the source code harder to read

kkraune:
yes I agree &lt;code&gt; does not look good.

kkraune:
looks good! merge at will. I like you added linebreaks, makes later diffs much easier to read

documentation/vespa-quick-start.html
jonmv: Could use —, I suppose, if that works for html.

jonmv:
I'll use <samp> instead of <code>; it's a monospace font without the red. 
I'll change &mdash; to — (it works).
I'll make sure the list looks good, w.r.t. vetical spacing between items. 
Then I'll merge :]

jobergum:
if you change  VESPA_SAMPLE_APPS you should update the linked blog search tutorial.

jonmv:
Good spot, Jo Kristian; I will do that. I was already going through that
document some weeks ago, but then the source got moved in the middle of it.
I'll finish that work when I get my hands on my modified copy (work machine
-- home today with sick child).

PR #37, opened 2017-08-24
Fix info flow document

kkraune:
@bratseth please merge

- set correct environment variables
- remove disused concepts
- deploy can be run anywhere
- no config files
- link to services, hosts.xml and tools where possible
- general re-indent and linebreaks
- link to more documents
- in general, much of this content is redundant
 - will consider this in later commits


bratseth:
Great

PR #58, opened 2017-08-31
Add note on load types and metrics

kkraune:
- moved from feeder doc where it did not belong

@jobergum please merge

### DIFF
--- a/documentation/reference/metrics-health-format.html
+++ b/documentation/reference/metrics-health-format.html
@@ -140,3 +140,29 @@ Each metric have one or more values set - valid values:
 </tr>
 </tbody>
 </table>
+
+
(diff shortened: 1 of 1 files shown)


jobergum:
Wrong document, the document talks about the state metrics api, the example with load type etc is from a different legacy metric interface.

kkraune:
hm, confused - metrics are available at state/v1/metrics, right, like 

https://api.vespa.corp.yahoo.com:4443/application/v4/tenant/dictionary_search/application/dictionary_vespa_app/environment/prod/region/ap-southeast-1/instance/default/service/searchnode-409hezp355p1oo1mypx7toha2/state/v1/metrics

or elsewhere?

PR #59, opened 2017-08-31
use YQL - this is the future query language

kkraune:
- yes, query=bad looks better, but better get used to YQL from the start
  - rest of docs will be ported to YQL
- bonus: do not use -s when checking for 200 OK, better user experience

@jobergum

### DIFF
--- a/documentation/vespa-quick-start.html
+++ b/documentation/vespa-quick-start.html
@@ -67,7 +67,7 @@ $ docker exec -it vespa bash
     <strong>Ensure the application is active:</strong><br>
     Wait for a valid <samp>200 OK</samp> response from:
 <pre>
-[root@vespa-container #/] curl -s --head http://localhost:8080/ApplicationStatus 
+[root@vespa-container #/] curl --head http://localhost:8080/ApplicationStatus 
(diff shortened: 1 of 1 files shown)


bratseth:
I think "easy to get started" trumps "better to get used to X" (because there are so many X'es).

documentation/vespa-quick-start.html
```diff
@@ -67,7 +67,7 @@
     <strong>Ensure the application is active:</strong><br>
     Wait for a valid <samp>200 OK</samp> response from:
 <pre>
-[root@vespa-container #/] curl -s --head http://localhost:8080/ApplicationStatus 
+[root@vespa-container #/] curl --head http://localhost:8080/ApplicationStatus
```
jobergum: Why this change? -s hides the progress stuff ?

kkraune: sorry my bad

kkraune:
&gt; I think "easy to get started" trumps "better to get used to X"

Agreed - I will instead evaluate the next steps section and make sure documents there use YQL natively

PR #66, opened 2017-09-01
Work outside docker container

frodelu:
@jobergum : My suggestion is to avoid exec'ing inside the Docker container. This example lets you work outside.


jobergum:
Yes, an alternative is to use http api's and export configuration server port and then use the REST configuration server api'. Then we won't need any docker exec at all but then there is a little more complex instructions.

frodelu:
Yes, that's an alternative, but I thought this was a decent compromise between the two approaches. 😄

kkraune:
it is a good compromise, the deployment URLs do not look good, this is better

jobergum:
Let us consolidate feedback but I think I agree as this as a reasonable
compromise. Maybe it all gets replaced with one big shell script next week.

PR #67, opened 2017-09-01
Kkraune/search api

kkraune:
@bratseth - this still needs more work to add a few more features, next PR


bratseth:
👍   whenever you see "QRS" it would be good to replace it by "container" or "search container".

PR #68, opened 2017-09-04
Use Search Container instead of QRS

kkraune:
- or just Container where fit
- did not find a safe way to fully automate, replaced by human
- formatting and linguistic fixes bonus while at it
- there are more docs, this is just a subset

@bratseth


bratseth:
💯

PR #69, opened 2017-09-05
fix feed commands, cannot have trailing space

kkraune:
- re-add 4G RAM requirement (ok to move this to a requirements section later)

@jobergum merge asap


jobergum:
👍

PR #70, opened 2017-09-05
merge commandline tools into one doc

kkraune:
- fix in/out-links
- fix broken content in the tool's docs
- makes the vespa TOC shorter, which was the intention of this

@ldalves this is a big diff, no need to review, just merge - we will take a second pass at the content


kkraune:
@ldalves ping

ldalves:
👍

PR #82, opened 2017-09-12
new documentation layout

ldalves:
@kkraune, please review and merge. Some footer links need to be confirmed. 

FYI: @frodelu


bratseth:
👍  👍

PR #88, opened 2017-09-13
fix html

kkraune:
@jobergum please merge


jobergum:
This is addressed in https://github.com/vespa-engine/documentation/pull/89, can we please have some patience with things we know is in progress?

PR #91, opened 2017-09-13
Remove "Welcome" link

kkraune:
- this has the same purpose as clicking the logo a few pixles above
  so I don't see a need for it
- maybe I am missing something, though, feel free to comment

@ldalves please merge

### DIFF
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -1,12 +1,5 @@
 
 <div class="col-sm-3 col-md-2 sidebar">
-	<ul class="nav nav-sidebar">
-		{% if page.url == '/' or page.url == '/documentation/' or page.url == '/documentation' %}
-			<li class="active"><a href="/documentation/">Welcome <span class="sr-only">(current)</span></a></li>
-		{% else %}
-			<li><a href="/documentation/">Welcome</a></li>
-		{% endif %}
-	</ul>
 	{% if site.data.sidebar.docs[0] %}
 		{% for item in site.data.sidebar.docs %}
 			<ul class="nav nav-sidebar">


ldalves:
Per the original design from Marco, we should have it... IMHO, the "looking and feel" is better starting with it on the left sidebar, but I'm ok if decide to remove it. 

@frodelu, what's your opinion here? 

I think what we should not have is the "/documentation" in the url, since "http://docs.vespa.ai/" and "http://docs.vespa.ai/documentation/" seems redundant.

bratseth:
> I think what we should not have is the "/documentation" in the url, since "http://docs.vespa.ai/" and "http://docs.vespa.ai/documentation/" seems redundant.

I don't think gh supports any url rewriting, so to avoid that we need to put all the documents in the root of https://github.com/vespa-engine/documentation ... I think that will be a bit messy.

ldalves:
Jekyll has a redirect plugin https://github.com/jekyll/jekyll-redirect-from that could be used, but in tests it shows "redirecting..." when accessing the target link.. then it's not nice. 

Agree that putting all the documents in the root, as we have them now, would be a bit messy... We would need to better organize/group them in "main" folders first...

bratseth:
> We would need to better organize/group them in "main" folders first...

Yes, we could move everything else to e.g site/
and then move the documents up.

kkraune:
decision: we will keep the Welcome, but please change the logo link to point to the frontpage, @ldalves

bratseth:
>  but please change the logo link to point to the frontpage

Just to ensure there is no misunderstanding: The logo should link to http://vespa.ai

ldalves:
done: https://github.com/vespa-engine/documentation/pull/93

PR #92, opened 2017-09-13
Cut less relevant docs from sidebar

kkraune:
- these are linked from elsewhere
- makes the sidebar more focused

@bratseth please merge

side note: given http://docs.vespa.ai/documentation/api.html I consider removing some of the API links from the sidebar, these docs are not much used (State API, Config API) - agree?

### DIFF
--- a/_data/sidebar.yml
+++ b/_data/sidebar.yml
@@ -44,9 +44,9 @@ docs:
         url: /documentation/jdisc/processing.html
       - page: Developing web service applications
         url: /documentation/developing-web-services.html
-      - page: Components for the JDisc Container
+      - page: Components for the Container
         url: /documentation/jdisc/container-components.html
-      - page: Component Types in the JDisc Container
+      - page: Component Types in the Container
         url: /documentation/component-types.html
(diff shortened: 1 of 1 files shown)


bratseth:
> I consider removing some of the API links from the sidebar, these docs are not much used (State API, Config API) - agree?

Yes.

bratseth:
Not completely convinced about the services sublinks, as those are used a lot, but I'll merge.

jobergum:
Could we move - title: RANKING up in the list? There is too much java stuff "DEVELOPING APPLICATIONS"  above the fold ?

kkraune:
&gt; are used a lot
in that case, your browser cache got the link ... :-)

kkraune:
I agree, @jobergum but in that case we will mix reference docs all over - but some times hard to see the distinction

jobergum:
maybe we get a different page rendered, but ranking with vespa is not a reference document but a great introduction and it should be above the fold.  Developing applications => Developing custom components.

kkraune:
ok - I'll drop by and we will make new PR with great improvements

PR #97, opened 2017-09-18
Kkaune/aws multi

kkraune:
@bergum it is only the first commit that needs review, makes the guide work again (I have tested it for a single-node setup)

The second commit makes the guide more read/writeable - please merge

@bjorncs FYI

### DIFF
--- a/documentation/vespa-quick-start-multinode-aws.html
+++ b/documentation/vespa-quick-start-multinode-aws.html
@@ -3,59 +3,72 @@
 title: "Multi-Node Quick Start: Install and run Vespa on AWS EC2"
 ---
 
-<p><strong>Prerequisites</strong>:
+<p>Prerequisites:
 <ul>
   <li>An AWS account </li>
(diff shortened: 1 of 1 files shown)


documentation/vespa-quick-start-multinode-aws.html
```diff
@@ -74,13 +74,13 @@ <h2 id="install-vespa">Install and run Vespa</h2>
 our case we chose to use <samp>ip-172-1-1-1.us-east-2.compute.internal</samp></p>
 <pre>
 $ for host in $(cat hosts.txt); do \
-ssh -f $host "sudo bash aws_bootstrap.sh $(localhost)"; done
+ssh -f $host "sudo bash aws_bootstrap.sh $(hostname)"; done
```
jobergum: I think this might just have to be 'ip-172-1-1-1.us-east-2.compute.internal' instead of $(hostname) to make it explicit.

PR #100, opened 2017-09-18
Jvenstad/blog search fixup

jonmv:
@lesters please review and merge :]


jonmv:
I'm also wondering whether we should tune down the `inline code` decoration a little — all the redness is a bit hurtful for tender eyes.

lesters:
👍

PR #112, opened 2017-09-20
 remove docker mem & add step for verify configuration server up

jobergum:
Removed docker memory entirely (unlimited is default). Added section on check configuration server.


kkraune:
> unlimited is default

is that so? on my home docker for mac, it says 2G

@bjorncs : you said 2G won't do?

PR #115, opened 2017-09-20
Remove contact document

bratseth:
@kkraune please review

This document was rather pointless, and we need a longer one anyway under frontpage


bratseth:
Try again CLA

kkraune:
super! this was a placeholder when eliminating the 100s of "mail vespa team" links in the historic doc, but not needed now - I assume this was the only reference left to it. one less file 🥇

bratseth:
Yes, that was the only ref

PR #118, opened 2017-09-21
Add link to tensor javadoc

lesters:
@kkraune Please review and merge.
@bratseth FYI.

I will add a few code examples here ASAP, but this is to remove the misunderstanding about whether the Tensor Java API itself is coming soon.


kkraune:
👍

PR #126, opened 2017-09-22
Update documentation toc, remove qrs mentions and add metrics doc to toc

jobergum:


kkraune:
Thanks. Generally I really dislike the use of upper case like in _Developing Web Service Applications_ - this is nor German language - _Developing web service applications_ and the likes are much better imho

I will merge anyhow to get the other changes in

jobergum:
I don't really care but the majority uses German casing so consistency was the issue for me, Feel free to change back but please be consistent.

PR #128, opened 2017-09-22
Make copy and paste behavoir for commands in quick start easier

smorgrav:
An alternative method would be to make the $ sign unselectable

### DIFF
--- a/css/docs.css
+++ b/css/docs.css
@@ -245,6 +245,13 @@ h3,
     padding-top: 0;
 }
 
+.force-select {  
+  -webkit-user-select: all;  /* Chrome 49+ */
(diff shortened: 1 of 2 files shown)


smorgrav:
Did a small survey and most people prefer the force select on the actual command. For obvious parts that would make sense to break up you can use multiple spans or make the $ sign unselectable as earlier noted.

kkraune:
ok with me, no strong opinions, merge at will (there are conflicts to resolve first)

jobergum:
@smorgrav please consider doing this for all 4 quick starts then, if you think this really helps on the user experience.

kkraune:
close or merge, @smorgrav ?

PR #141, opened 2017-09-25
Bratseth/doc improvements

bratseth:
@kkraune please review


bratseth:
These documents are linked from the features section on frontpage so we should show them some extra special love.

kkraune:
Thanks! 💯

jobergum:
Great stuff!

PR #149, opened 2017-09-27
Remove id attr on anchor to fix formatting

kkraune:
- not sure why the id was there
- copy seems to work without the id

ref slack feedback this morning

@lesters please review and merge

@ldalves FYI - not sure why the skew in rendering, maybe some magic due to the header


lesters:
@kkraune There is a issue which @ldalves is working on, where all items that lists with an id field will be rendered wrong. The problem is the following CSS:

https://github.com/vespa-engine/documentation/blob/master/css/docs.css#L248-L255

I think @ldalves had a general solution now?

lesters:
The `id` field is to be able to link directly to the example, but it is not crucial.

ldalves:
PR #150

PR #150, opened 2017-09-27
fix issue with toc and lists / links

ldalves:
@lesters, as far I can see, this should fix the issue. Could you please help verify? I tested adding the `{toc}` int he blog recommendation tutorial and it seems ok as well. 

@kkraune, FYI.


lesters:
Let's try it.

PR #151, opened 2017-09-27
Don't urldecode the urlencoded example.

alexanderkjeldaas:

PR #152, opened 2017-09-27
Don't urldecode string which is supposed to show as encoded as an exa…

jobergum:
…mple

https://github.com/vespa-engine/documentation/pull/151


kkraune:
there are many more in the file, should fix all

jobergum:
Note that this is the only example where you actually want to display the urlencoded version, other examples is as far as I can tell just fine @kkraune.

PR #155, opened 2017-09-27
Fix mobile display by adding viewport meta tag

jdomingu:
On mobile, pages are not displaying according to the media query rules in the CSS. Instead, pages are being scaled down because the viewport isn't set. (I'm seeing this in Google Chrome both on Android and in the device mode in Chrome developer tools. This likely reproduces on iOS as well, though someone else would have to confirm.)

Adding the viewport is recommended in the [Bootstrap docs](http://getbootstrap.com/2.3.2/scaffolding.html#responsive) and is pretty standard:

* [Using the viewport meta tag to control layout on mobile browsers](https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag) in the MDN Web Docs
* [Configuring the Viewport](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/UsingtheViewport/UsingtheViewport.html) in the Safari Web Content Guide

Congrats on open sourcing Vespa! The docs look great.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -1,4 +1,5 @@
 <meta charset="utf-8" />
+<meta name="viewport" content="width=device-width, initial-scale=1">
 <title>{{ page.title | escape }}</title>
 <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" integrity="sha384-1q8mTJOASx8j1Au+a5WDVnPi2lkFfwwEAa8hDDdjZlpLegxhjVME1fgjWPGmkzs7" crossorigin="anonymous">
 <link rel="stylesheet" href="{{site.github.url}}/css/font-awesome/css/font-awesome.min.css">


yahoocla:

jobergum:
Thanks a lot for your PR! Would you mind signing the CLA quoted by the bot above?

jdomingu:
Signed. Thanks!

frodelu:
@ldalves Please review.

ldalves:
thanks, @jdomingu!

ldalves:
I see the sidebar stopped working for some reason and the menu options are not responding as they should... sorry we'll have to revert and investigate a bit more about this.

jdomingu:
@ldalves @frodelu Could you elaborate more on the sidemenu issues (and the browser that you're using)? Not seeing problems with a local jekyll build. I do see that the navbar menu is broken on mobile--that seems unrelated to the change though. From what I can tell, the menu issues are being caused by missing files. The menu collapse functionality depends on JavaScript in the bootstrap.min.js file, which for some reason isn't being included on any of the pages. I would put this in your _includes/header.html:

```
<script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.4/jquery.min.js"></script>
<script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js"></script>
```

Note that bootstrap isn't compatible with jquery 3--you'll need to switch to a different version of jquery as above.

Optionally, you might want to do some cleanup. Most of the files in the css/ dir aren't being used since the header links to files from a CDN. The mobile menu (and search bar in particular) also have some style issues that would be nice to fix.

ldalves:
@jdomingu, I have done the clean-up and fix the navbar on this PR https://github.com/vespa-engine/documentation/pull/160.

The issue with the sidebar is that we miss it when using the _"viewport"_... We need to have a _toggle_ button there as well, but since it's fixed, not sure yet how to solve this. 

What are the style issues that would be nice to fix? 

Thanks for your feedback.

jdomingu:
Awesome! Glad to see the navbar got fixed. The style issues are mostly
suggestions--here are some things I noticed though:

* The main content becomes somewhat hard to read on large screens because
it always spans the whole screen. I would add a max-width: 75em; style to
.main (or something similar).
* On smaller screen sizes, the search input in the menu is very small. It
would take some work to get this to be full-width and yet keep the submit
button aligned properly.
* Part of the sidebar scrollbar is obscured by the header--it could use a
small increase in the top margin.

Love that you all are so responsive and continuing to make improvements :)

PR #170, opened 2017-10-03
Eliminate one multinode install doc

kkraune:
- the aws doc got all info from the old multinode install doc - remove the latter
- fixed links
- bonus: the jdisc doc got a tiny unrelated overhaul

@frodelu please merge

@jobergum FYI and thanks for the nice aws multinode doc


kkraune:
@frodelu ping

PR #171, opened 2017-10-03
remove references to non-existing scripts

havardpe:
also recommend not saving everything to disk

@kkraune please review


kkraune:
Great - thanks!

PR #177, opened 2017-10-12
Doc for plugin upgrade

freva:
I have another paragraph on locking, but I can't think of any good usecase, and its not the easiest to use so I don't think it should be public anyway.


kkraune:
sorry this got lost - looks good! merging

PR #183, opened 2017-10-18
Add link to sample app

jobergum:
Add link and use 'setup' instead of settings as that is what the text box is named in the playground.


jobergum:
cc @lesters

PR #184, opened 2017-10-18
Fix bundle name

jobergum:
@lesters I'm really uncertain about the flow in this document, why don't we just point to the blog-recommendation sample app and quote less code in the document itself?


lesters:
@jobergum That's a good idea, I'll take a stab at this ASAP.

PR #186, opened 2017-10-19
bundle JNI

kkraune:
@bratseth maybe as simple as this?

@gjoranv FYI

### DIFF
--- a/documentation/jdisc/developing-osgi-bundles.html
+++ b/documentation/jdisc/developing-osgi-bundles.html
@@ -142,46 +142,8 @@ Import-Package: org.osgi.framework;version="1.3.0"
 
 <h2 id="depending-on-JNI-libraries">Depending on JNI Libraries</h2>
 <p>
-When a library uses the Java Native Interface (JNI), it is crucial that a
-compatible version of the native library is installed on the host.
-This means that even if the Java library is packaged as an OSGi bundle,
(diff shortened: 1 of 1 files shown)


bratseth:
great

gjoranv:
Do you have a working example? This PR removes a lot of info.

bratseth:
Here's a working example: https://git.corp.yahoo.com/ml/vespa-tensor-serving
But it does not contain anything special, just a dependency which happen to contain an .so.

jobergum:
Agree with @gjoranv, this page had a lot of potential useful information on JNI and now it contains nothing, no examples and leaves a lot to be guessed by the developer. 👎

kkraune:
so, I'm no expert but according to @bratseth the previous would not work anyway? maybe I got it wrong. In that case - feel free to add back the missing

gjoranv:
@bratseth, why didn't this work? I'm quite sure we've assisted customers with that solution earlier. It's the only way a user can make a library available on the classpath. It's not an ideal solution, but since native libs can only be loaded once it's sometimes the only way.

bratseth:
It could work, but not on hosted, and it is misleading - in most cases this is the wrong thing to do.
I'll add back the information about the classpath.

bratseth:
https://github.com/vespa-engine/documentation/pull/187

PR #187, opened 2017-10-20
Add classpath alternative

bratseth:
@gjoranv please review

@jobergum fyi

### DIFF
--- a/documentation/jdisc/developing-osgi-bundles.html
+++ b/documentation/jdisc/developing-osgi-bundles.html
@@ -146,7 +146,7 @@ This section details alternatives for using native code in the container.
 
 <h3>OSGi bundles containing native code</h3>
 
-<p>OSGi jars may contain .so files, which will can be loaded in the standard way from Java
+<p>OSGi jars may contain .so files, which can be loaded in the standard way from Java
 code in the bundle. Note that since only one instance of an .so can be loaded at any time,
 it is not possible to hot swap a jar containing .so files - when such jars are changed
 the <a href="container-components.html#JNI-requires-restart">new configuration will not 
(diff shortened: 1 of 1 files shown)


documentation/jdisc/developing-osgi-bundles.html
```diff
@@ -141,9 +141,57 @@ <h2 id="typical-bundle">What is an OSGi bundle?</h2>
...
-To use JNI libraries, add the .so file(s) to a bundle and depend on this as any other dependency.
-Tensorflow is an example bundle with JNI.
-Read more on <a href="container-components.html#JNI-requires-restart">
-JNI and deployments</a>, as restarts are required when the bundle with native files change.
+
+This section details alternatives for using native code in the container.
+
+<h3>OSGi bundles containing native code</h3>
+
+<p>OSGi jars may contain .so files, which will can be loaded in the standard way from Java
```
gjoranv: s/will can/can/

documentation/jdisc/developing-osgi-bundles.html
```diff
@@ -141,9 +141,57 @@ <h2 id="typical-bundle">What is an OSGi bundle?</h2>
...
+code in the bundle. Note that since only one instance of an .so can be loaded at any time,
```
gjoranv: Minor but interesting: Should we read '.so' as "ess oh" or "dot ess oh"?  If the former, then 'an' is correct'. Otherwise it should of course be 'a'.

gjoranv:
One more thing. I assume that bundling the .so only works when the native code is your own, and is inside the same bundle? If so, maybe we should emphasize that the origin of the native code decides which method is possible to use.

documentation/jdisc/developing-osgi-bundles.html
bratseth: Fixed.

documentation/jdisc/developing-osgi-bundles.html
bratseth: Imho it's "ess of".

bratseth:
It doesn't have to be "your own", and I think i already emphasise that the Java code using it must be in the same bundle

Of course it can't rely on binaries which are not in the bundle, but I think that's too obvious to mention?

PR #188, opened 2017-10-20
Config definitions moved to another location.

aressem:
@hmusum


hmusum:
Merged this since it is correct today too (due to symlink)

PR #190, opened 2017-10-20
Document id strings can only contain text characters.

geirst:
@baldersheim or @jobergum please review


geirst:
@baldersheim @jobergum ping

PR #192, opened 2017-10-24
Add generic multinode install info and make some clarifications

bratseth:
@jobergum please review, or @kkraune 

We were missing a generic multinode install now. 
Added one, and clarified some things which has come up in meetings with external users.

### DIFF
--- a/documentation/multinode-systems.html
+++ b/documentation/multinode-systems.html
@@ -1,0 +1,63 @@
+---
+# Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+title: "Multinode systems"
+---
+
+<p>A Vespa <i>system</i> consists of one or more stateless and stateful clusters configured by one application package.
+A Vespa system is configured and managed through an admin cluster as shown below.</p>
+
+<img src="img/overview/vespa-overview.png"/>
+
+<p>All nodes of a Vespa system has the same software installed. Which processes are started on each node and
+how they are configured is determined by the admin cluster from the specification given in the 
+services.xml file in the application package.</p>
+
+
+<h2 id="making-changes-to-live-systems">Making changes to live systems</h2>
+
+<p>To change the system, deploy the changed application to the admin cluster.
+The admin cluster will automatically change the participating nodes as necessary.
+It is safe to do this while serving live query and write traffic. 
+In rare cases the admin cluster will tell you that some processes must be restarted
+to make the change effective. To avoid query or write traffic disruption, such
+restarts must be done on one node at the time,
+and you must wait until the node is fully up before restarting the next node.</p>
+
+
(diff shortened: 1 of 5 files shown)


documentation/multinode-systems.html
```diff
@@ -0,0 +1,63 @@
...
+
+<h2 id="creating-a-multinode-system-from-a-sample-application">Creating a multinode system from a sample application</h2>
+
+<p>To create a fully functional production ready multinode system from a single-node sample application,
+follow these steps:
+
+<ul>
+<li>Add an explicit admin cluster to your services.xml file:
+<pre>
+&lt;services version=&quot;1.0&quot;&gt;
```
kkraune: oh I dislike the &amp;quot;  can you not just use  regular " ? ;-)

documentation/multinode-systems.html
```diff
@@ -0,0 +1,63 @@
...
+<li>Add these nodes to your container and content clusters as desired by adding more <code>node</code> tags in services.xml (to avoid traffic disruption on restarts or node failure you need at least two nodes in each cluster). You can add the same nodes
+to multiple clusters if you like.
+</ul>
+
+With this, your application is fully production ready!
+
+<h2 id="log-collection">Log collection</h2>
+
+<p>Logs are automatically collected from all nodes to the admin cluster nodes in real time.
+To view log messages from the system, run <code>vespa-logfmt</code> on an admin node.
```
kkraune: i _think_ the log archive is set up on one node only, so one of the admin servers is probably right - not all of them - or am I wrong?

documentation/vespa-quick-start-centos.html
```diff
@@ -128,22 +128,24 @@ <h2 id="next-steps">Next Steps</h2>
...
       sample applications</a>
-  </li><li>
+  <li>
     How to <a href="securing-your-vespa-installation.html">secure a Vespa installation</a>
-  </li><li>
-    Try the <a href="vespa-quick-start-multinode-aws.html">AWS EC2 multinode quick start</a>
+  <li>
+    To run on AWS, see the <a href="vespa-quick-start-multinode-aws.html">AWS EC2 multinode quick start</a>
     and <a href="vespa-quick-start-multinode-aws-ecs.html">AWS ECS multinode quick start</a>
-  </li>
```
kkraune: i find my xcode editor some times confused in indenting html without closing tags, that is why I normally keep them

kkraune:
This is a great addition! I will merge and later re-read it and possibly link to a few other docs wrt changes and restarts (somewhere in the search def doc)

documentation/multinode-systems.html
bratseth: I agree with you. I just use my editor to convert to entities automatically. I'll try to avoid them in the future.

documentation/multinode-systems.html
bratseth: Yes only the "admin server", not the "config servers" - I didn't want to get into that difference but now it's just confusing. PR incoming.

PR #194, opened 2017-10-25
Clear the steps from the previous run when having multiple pages to e…

aressem:
…xecute.

This will prevent steps accumulating in an unwanted manner.


lesters:
LGTM.

PR #196, opened 2017-10-25
Print source location when printing execution script

lesters:
@aressem Please review.

One problem now is that if one page fails, it will not continue to test remaining pages. I'll address this shortly.


aressem:
👍

PR #204, opened 2017-10-27
Added missing step. Encode &

aressem:
@jobergum  or @kkraune 

One step was missing from the previous merge. Fixed the encoding.


kkraune:
👍

PR #208, opened 2017-11-01
Lesters/doctest in pseudo terminal

lesters:
@aressem Please review.

This will run a pseudo terminal for each file so that the result of shell commands like "export" and "cd" will persist.


aressem:
Nice 👍

PR #209, opened 2017-11-02
Update file distribution doc to current state

hmusum:
* Deprecate filedistribution setting that will not work when
  we start using new implementation


kkraune:
Thanks!

PR #211, opened 2017-11-07
Kkraune/doc distribution

kkraune:
keep global docs in reference docs, to not confuse users. less bucket split details, use table for readability

@hmusum please merge


hmusum:
CLA check not working, maybe this comment helps?

PR #214, opened 2017-11-13
Clean up blog search tutorial

kkraune:
- remove simple query language
- make more consistent
- less text

@lesters please merge @bratseth FYI


lesters:
👍

PR #216, opened 2017-11-15
Add documentation of wantedDocumentCount visiting parameter

vekterli:
@kkraune please review. Can reduce level of waffling about edge cases if desired.


kkraune:
gr8 - thanks!

PR #218, opened 2017-11-17
Fix tutorial

kkraune:
- no random wait for pig feed to speed up
- must build and deploy before feeding
- make pig command work, seems it prefers -D first
- add deploy details

@lesters please merge


lesters:
👍

PR #220, opened 2017-11-21
Tutorial fixes

kkraune:
- reordered
- link to container intro, which got a bonus refactoring - do not use jdisc
- use current code - not sure whether to keep code or link to it
- added more output to make sure everything works

@lesters sorry for big diff - need another pass to test the rest, commit now to keep size down


lesters:
👍

PR #224, opened 2017-11-24
Use 'container' instead of 'container-dev' in pom.

gjoranv:


documentation/jdisc/processing.html
```diff
@@ -23,8 +23,8 @@ <h2 id="using_processing">Using processing</h2>
 <pre>
 &lt;dependency&gt;
   &lt;groupId&gt;com.yahoo.vespa&lt;/groupId&gt;
-  &lt;artifactId&gt;container-dev&lt;/artifactId&gt;
-  &lt;version&gt;6.9.0&lt;/version&gt; &lt;!-- Note: use latest version available --&gt;
+  &lt;artifactId&gt;container&lt;/artifactId&gt;
+  &lt;version&gt;6.158.42&lt;/version&gt; &lt;!-- Note: use latest version available --&gt;
```
bjorncs: Consider replacing a number with a dummy placeholder version that is guaranteed to fail.

PR #225, opened 2017-11-24
Refactor document

kkraune:
- less language
- linbreaks, lists
- only real change is adding a mixed tensor     tensor(x[],y{})
 - removing the note that is was not valid

@lesters please merge


lesters:
👍

PR #226, opened 2017-11-27
Improve tensor guide

kkraune:
- mixed tensor
- less language
- example includes a query (searcher not strictly needed anymore)
- did not fix ranking section and below, yet

@lesters please merge


lesters:
👍

PR #227, opened 2017-11-28
Clarify how imported fields are used in document summary.

geirst:
@jobergum please review


bratseth:
I couldn't find a system test of this when looking earlier, do we have it?

The documentation also says that all field settings except fast-search is carried over to the imported field. This seems to contradict that.

vekterli:
We have an internal test for imported attributes in summaries in `ParentChildAttributeSearchTest / document_summary_containing_imported_fields`, but it uses a custom summary and _not_ the default summary.

PR #228, opened 2017-11-30
improve tensor guide

kkraune:
- better doc on constant tensors
- not sure if 'constants' is mandated in app package, though
- fixed links to sample app, offset was wrong
- primarily changes to language and format - now more compact
- better linkage to/from

@lesters please merge. @hmusum FYI


lesters:
👍

PR #231, opened 2017-12-17
searchable copies min

kkraune:
- ref https://github.com/vespa-engine/vespa/blob/master/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/RedundancyBuilder.java#L31


kkraune:
@bratseth as discussed, please merge. @vekterli FYI

PR #234, opened 2018-01-15
Bergum/caches in vespa

jobergum:
New overview of caches in Vespa. Brought back cachesize in local providers. Please review @bratseth


jobergum:
FYI @baldersheim

PR #246, opened 2018-02-08
Fix rendering issue

yngveaasheim:
Use of 'id' inside of <tr > tag causes columns to be skewed.

@kkraune Please review and merge.


kkraune:
how odd - we need to have an ID in order to link to the element, can we put the id on the td instead?

kkraune:
👍  thanks for fixing the rendering

PR #251, opened 2018-03-05
WIP: Translate introductory documentation to Japanese

remoriok:
Adding some introductory documentation (from "About the Documentation" to "Quick Start") into `ja` directory.

An issue is how linking to the missing (non-translated) documentation should be done. I guess the whole documentation need to be published under `ja` directory for navigational consistency, but placing verbatim copies of the original documentation seems not reasonable to me.

Any ideas?


yahoocla:

remoriok:
FYI: I have signed the CLA.

ja/features.html
```diff
@@ -0,0 +1,111 @@
+---
+# Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
```
kkraune: 2018

kkraune:
Thanks, this is great! Yes, we should not copy the english documents. The best solution would be to link as if the Japanese document exists, and redirect to the English version if it did not. I will merge this so we can start experimenting with the linking.

BTW: I am in Tokyo coming week for the Vespa meetup, we should meet then :-)

remoriok:
Thanks! I'm looking forward to seeing you in Tokyo.

PR #254, opened 2018-03-09
Add anchor

bratseth:


bratseth:
@lesters please review whenever is convenient for you. No rush  :-)

PR #257, opened 2018-03-13
test redirects

kkraune:
inspired by https://help.github.com/articles/redirects-on-github-pages/ I suggest we use the _jekyll-redirect-from_ as a simple way to redirect to the English documents for documents that are not translated from say Japanese.

it is not a super scalable way, as each document will have an entry per language - on the other hand, it is not a practical problem short term as it is easy to add such an entry for all documents that are not translated from Japanese yet.

with this, once a document is translated, one must also remember to remove the redirect entry in the English doc (it loops if not, when tested on my laptop).

you like, @frodelu  and @lesters ? please be careful when merging, as I am not sure if the jekyll-redirect-from plugin is active on github, so you should merge, test and roll back if documents do not load (and today is not a good day to break Vespa docs with the Tokyo meetup)

@remoriok FYI we can hopefully fix so there are no dead links in Japanese docs

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -21,5 +21,8 @@ search:
     doctype: "open"
     endpoint: "http://doc-search.vespa.ai/vespa_documentation_search_proxy"
 
+plugins:
+  - jekyll-redirect-from
+
 # Build settings
 markdown: kramdown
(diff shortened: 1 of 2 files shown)


andreer:
This looks neat but forgotten @kkraune, shall we merge it and see?

kkraune:
yes forgot this, let see!

PR #263, opened 2018-03-21
Add deployment notes for enabling multiple bucket spaces

vekterli:
@toregge FYI

In particular, nodes must be restarted to enable conditional protocol features
that are currently set upon process startup.


kkraune:
👍

PR #290, opened 2018-04-26
Update reference doc for parent-child to include grandparent example …

geirst:
…as that is supported now.

@vekterli and @toregge please review


geirst:
Issue https://github.com/vespa-engine/vespa/issues/3339

PR #297, opened 2018-06-06
Bbratseth/quickstart improvements

bratseth:
@jobergum please review

### DIFF
--- a/documentation/vespa-quick-start-centos.html
+++ b/documentation/vespa-quick-start-centos.html
@@ -124,6 +124,13 @@ $ vagrant destroy
 
 </ol>
 
+
+
+<h2 id="troubleshooting">Troubleshooting</h2>
+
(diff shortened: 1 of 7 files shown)


bratseth:
@lesters it seems that the documentation rep is doing the testing of the sample-apps repo such that PRs on documentation fails when sample-apps are wrong, but sample-apps does not?

lesters:
@bratseth Well, the sample-apps testing is just building and running unit tests, while the doc testing is more of a integration / system test.

bratseth:
But don't you think the unit tests of the sample-apps repo should be in the sample-apps repo?

bratseth:
Ok, maybe it's fine. Looks to me that it is only testing the sample-apps but didn't look to carefully.

lesters:
Yes, well, that is mostly true 😄 : https://github.com/vespa-engine/documentation/blob/master/test/_test_config.yml

Currently, only quick start is tested in addition to sample apps. I'm in the process of adding tutorials as well to that list.

PR #309, opened 2018-06-22
Freshen up documentation for Document JSON updates

vekterli:

Restructures doc to be oriented more around the type of field being
updated, since that's likely to be what people are looking for.

### DIFF
--- a/documentation/reference/document-json-update-format.html
+++ b/documentation/reference/document-json-update-format.html
@@ -138,7 +138,7 @@ Individual struct fields may be updated using <a href="#fieldpath">field path</a
 Individual map entries may be updated using <a href="document-field-path.html">field path</a> syntax.
 </p>
 <p>
-Assigning an element to a key in a map will insert the key/value mapping if does not already exist, or
+Assigning an element to a key in a map will insert the key/value mapping if it does not already exist, or
 overwrite it with the newly assigned value if it does exist.
 </p>
 <h5>Map to primitive value</h5>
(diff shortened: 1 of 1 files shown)


documentation/reference/document-json-update-format.html
```diff
@@ -74,90 +89,205 @@ <h3 id="assign">assign</h3>
...
+  field last_name type string {}
+}
+field contact type person {
+  indexing: summary
+}
+</pre>
 <pre>
 {
-    "update": "id:music:music::example2",
+    "update": "id:mynamespace:workers::example",
```
kkraune: yes! good practice to say mynamespace wherever, users are confused about music:music :-)

vekterli: I still get confused myself, to be honest 😬 Principle of Least Surprise™️ reigns supreme

documentation/reference/document-json-update-format.html
```diff
@@ -74,90 +89,205 @@ <h3 id="assign">assign</h3>
...
+<p>
+<strong>Note:</strong> this does not work for <em>attribute</em> struct fields.
+</p>
+
+<h3 id="assign-map-field">Map field</h3>
+<p>
+Individual map entries may be updated using <a href="document-field-path.html">field path</a> syntax.
+</p>
+<p>
+Assigning an element to a key in a map will insert the key/value mapping if does not already exist, or
```
geirst: Typo? "...mapping if does not..." -> "...mapping if it does not"?

vekterli: Fixed

PR #314, opened 2018-06-27
Add testing of blog search tutorial

lesters:


lesters:
@kkraune FYI
@aressem FYI: adds ability to write files during testing, here to write services.xml, hosts.xml and blog_post.sd

PR #315, opened 2018-06-28
Add reference doc on struct and map attribute support for indexed sea…

geirst:
…rch.

@vekterli please review
@kkraune FYI

### DIFF
--- a/documentation/reference/search-definitions-reference.html
+++ b/documentation/reference/search-definitions-reference.html
@@ -240,13 +240,16 @@ struct [name] {
 }
 </pre>
 The struct name should not contain any underscores.
-</p><p>
-Please note that struct types is not supported by indexed search and
-the behavior of operations over struct fields is undefined.
-<!--  ToDo Rewrite: struct fields in summary is OK - i.e. use structs in document store -->
-Searching in struct elements is  supported by
-<a href="../streaming-search.html">streaming search</a>.
-</p><p>
+</p>
+<p>
+Note that struct types are supported differently in indexed search and
+<a href="../streaming-search.html">streaming search</a> mode.
+Take a look at
+<a href="#type:struct">struct type</a>,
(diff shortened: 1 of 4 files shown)


documentation/reference/search-definitions-reference.html
```diff
@@ -2303,19 +2322,66 @@ <h2 id="field_types">Field types</h2>
...
+    </tr>
+  </tbody>
+</table>
+</td></tr>
+
+<tr><th id="type:array-struct">array&lt;struct-type&gt;</th>
+<td>
+Use to create an array field of the given <a href="#struct">struct type</a>.
+The struct type must be defined separately.
+Restrictions:
```
vekterli: Should we move this _below_ the example to avoid breaking the flow of "here's what you do" and the example showing it? This also improves the proximity of "See table below for supported cases", since the table will be just below the list.

Also consider adding a `<p></p>` around Restrictions to ensure it gets a line of its own to strut around on (unless this looks like crap inside a table).

documentation/reference/search-definitions-reference.html
```diff
@@ -2400,67 +2466,105 @@ <h2 id="field_types">Field types</h2>
...
+field identities type map&lt;string, person&gt; {
+  indexing: summary
+  struct-field key { indexing: attribute }
+  struct-field value.last_name { indexing: attribute }
+}
+</pre>
+The entire <em>identities</em> field is part of document summary,
+and the struct fields <em>key</em> and <em>value.last_name</em> are attributes available for searching using the
+<a href="../query-language.html#same-element">sameElement</a> operator.
+The next example shows a map of primitive types, where the <em>key</em> and <em>value</em> struct fields are specified as attributes:
```
vekterli: Consider adding a paragraph break before this sentence since it's a slight change of context

geirst:
@vekterli PTAL (all comments addressed :) )

PR #347, opened 2018-08-15
Added page comparing Vespa and Elasticsearch

olaaun:


olaaun:
Commenting for license re-evaluation

PR #348, opened 2018-08-15
Querying Vespa brush up

ollivir:
- add simplified process image to the page
- clean up some of the text
- add to ToC


olaaun:
👍

PR #357, opened 2018-09-12
Bratseth/stateless evaluation doc

bratseth:
@lesters please review


bratseth:
I did write it out, thanks.
And yes, we should have a sample app.

PR #362, opened 2018-09-13
Extend field path update doc with which fields it supports.

geirst:
@toregge please review

### DIFF
--- a/documentation/document-api-guide.html
+++ b/documentation/document-api-guide.html
@@ -183,11 +183,19 @@ which is a <a href="reference/document-select-language.html">document selection 
 The <em>where clause</em> decides whether the document is to be changed at all,
 and also sets variables to use in the field path
 when deciding which parts of the document to change.
+</p>
+<p>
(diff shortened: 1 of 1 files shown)


documentation/document-api-guide.html
```diff
@@ -183,6 +183,11 @@ <h3 id="field-path-updates">Field Path Updates</h3>
 The <em>where clause</em> decides whether the document is to be changed at all,
 and also sets variables to use in the field path
 when deciding which parts of the document to change.
+Field path updates are only supported on non-attribute
+<a href="reference/search-definitions-reference.html#field">fields</a>,
+<a href="reference/search-definitions-reference.html#index">index</a> fields,
```
toregge: If a field is both an indexed field and an attribute, then the document is updated in the document store, the index is updated, but the attribute is not updated. Thus you can get old values in document summary requests and old values being used in ranking and grouping.

geirst:
@toregge PTAL

PR #363, opened 2018-09-18
render better

kkraune:
@bratseth


kkraune:
yes, got some odd grey squares in chrome - gone now

PR #364, opened 2018-09-19
'function', not 'macro'

bratseth:
@lesters please review but do not merge. I'll hold that off until a release with this is out.


bratseth:
Perhaps @kkraune too

lesters:
LGTM!

PR #365, opened 2018-09-19
clarify

kkraune:
@vekterli please merge

### DIFF
--- a/documentation/reference/services-content.html
+++ b/documentation/reference/services-content.html
@@ -284,7 +284,9 @@ Configure this cluster and chain as the indexing chain in the content cluster - 
 <h2 id="redundancy">redundancy</h2>
 <p>
 Contained in <code><a href="#content">content</a></code>.
-Defines the total number of copies of each piece of data the cluster will maintain to avoid data loss.
+Defines the total number of copies of each piece of data the cluster will maintain to avoid data loss
+- within a <a href="#group">group</a>.
+If <em>group</em> is not configured, this equals one group, hence all nodes.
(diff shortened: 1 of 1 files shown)


documentation/reference/services-content.html
```diff
@@ -284,7 +284,9 @@ <h2 id="document-processing">document-processing</h2>
 <h2 id="redundancy">redundancy</h2>
 <p>
 Contained in <code><a href="#content">content</a></code>.
-Defines the total number of copies of each piece of data the cluster will maintain to avoid data loss.
+Defines the total number of copies of each piece of data the cluster will maintain to avoid data loss
+- within a <a href="#group">group</a>.
+If <em>group</em> is not configured, this equals one group, hence all nodes.
```
vekterli: Doesn't group-aware auto scaling of redundancy only apply iff you're running in a hosted setting?

documentation/reference/services-content.html
```diff
@@ -524,6 +526,8 @@ <h2 id="searchable-copies">searchable-copies</h2>
 The benefit is using less memory, trading off availability during transitions.
 Refer to <a href="../proton.html#bucket-move">bucket move</a>.
 </p><p>
+<em>searchable-copies</em> applies within a <a href="#group">group</a>, like <em>redundancy</em>.
+</p><p>
```
vekterli: Same question as above

documentation/reference/services-content.html
kkraune: don't know ...

bratseth: Yes, "within a group" is only true on hosted Vespa.

PR #381, opened 2018-10-11
Fix typo in release notes

bjorncs:


bjorncs:
Assuming the documentation was referring to the new parameter for `vespa-get` / `vespa-visit`.

PR #392, opened 2018-10-30
Bratseth/various improvements

bratseth:


bratseth:
@kkraune please review

PR #394, opened 2018-11-12
add default for flush-on-shutdown

kkraune:
I don't know if the default is true or false - could not see from https://github.com/vespa-engine/vespa/blob/a902861604ddbaa45be0078a733add8f5009a30e/config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java#L84 nor https://github.com/vespa-engine/vespa/blob/49a532f515d9ce217def27443b5bb9faa6410124/config-model/src/main/resources/schema/content.rnc#L164

### DIFF
--- a/documentation/reference/services-content.html
+++ b/documentation/reference/services-content.html
@@ -538,8 +538,7 @@ Note that this is only useful if <code>searchable-copies</code> is less than <co
 
 <h2 id="flush-on-shutdown">flush-on-shutdown</h2>
 <p>
-Contained in <code><a href="#proton">proton</a></code>.
-Defult value is false.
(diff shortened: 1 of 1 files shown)


documentation/reference/services-content.html
```diff
@@ -539,6 +539,7 @@ <h2 id="searchable-copies">searchable-copies</h2>
 <h2 id="flush-on-shutdown">flush-on-shutdown</h2>
 <p>
 Contained in <code><a href="#proton">proton</a></code>.
+Defult value is false.
```
geirst: Typo: Defult -> Default

kkraune:
Is false correct you think?

geirst:
I will test to verify.

PR #397, opened 2018-11-20
Specify that document-api does not imply search on 7

bratseth:
@arnej27959 please review


aressem:
It seems like this broke the boolean search example:
https://travis-ci.org/vespa-engine/documentation/builds/458426052

PR #402, opened 2018-11-30
Remove experimental enable-multiple-bucket-spaces

bjormel:
This feature is enabled by default on Vespa 6.245 and beyond. You don't need to set this flag unless you are on an older version.

Ref https://github.com/vespa-engine/vespa/pull/7753


bjormel:
@geirst 
@vekterli

PR #405, opened 2018-12-02
Update terminology

bratseth:
@ollivir please review


ollivir:
LGTM. There should probably be some more words about what the numbers mean (introduction order, not dist-key/group id), but that doesn't affect this change.

PR #411, opened 2018-12-10
add time.date

kkraune:
- probably left out by mistake


kkraune:
ref https://github.com/vespa-engine/vespa/blob/278b2cb304fc1c98691e0b36de42a31c9128172d/container-search/src/main/javacc/com/yahoo/search/grouping/request/parser/GroupingParser.jj#L151

PR #412, opened 2018-12-10
move examples from reference to guide

kkraune:


documentation/reference/grouping-syntax.html
```diff
@@ -381,247 +380,6 @@ <h2 id="expression">Expressions</h2>
 
 
 
-    <h2 id="examples">Examples</h2>
```
bjorncs: The example section contains a more detailed description of some of the operators. I agree on moving the examples, but we should consider keeping the operator descriptions in this document.

bjorncs: E.g first section of _Counting unique groups_

kkraune: agreed! I will add back to reference guide what are not strictly examples, easier to do not as the reference is more like a reference

documentation/reference/grouping-syntax.html
```diff
@@ -381,247 +380,6 @@ <h2 id="expression">Expressions</h2>
...
-    <pre>/search/?yql=select (&hellip;) |
-             all(group(predefined((now() - a) / (60 * 60 * 24),
-                                  bucket(0,1), bucket(1,2),
-                                  bucket(3,7), bucket(8,31))) each(output(count())
-                 all(max(2) each(output(summary())))
-                     all(group((now() - a) / (60 * 60 * 24)) each(output(count())
-                         all(max(2) each(output(summary())))))));</pre>
-
-
-    <h3 id="counting-unique-groups">Counting unique groups</h3>
```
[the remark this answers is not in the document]
kkraune: fixed

documentation/reference/grouping-syntax.html
kkraune: I will do that in a separate PR

PR #419, opened 2018-12-12
move content to right place

kkraune:
the gist of this is to move content to one place and make it coherent

a few minor additions buried in all this, not critical to review, manually tested by me - and the reference should be reviewed later after addressing all ToDos (added a few more of those...)

assigning to @bratseth as @bjorncs is busy today


kkraune:
I assume array.at is an Accessor expression, added there

kkraune:
@bjorncs please review if time

PR #420, opened 2018-12-12
Document limited support for importing complex fields:

toregge:
array of struct, map of struct, map, position, array of position.

@geirst: please review


kkraune:
ref #7206

PR #428, opened 2018-12-19
Clarify attribute vs. index

bratseth:
@kkraune please review

No wonder many people have been confused about this ...


kkraune:
you are awarded the 'help-future-users' prize for Dec 2018. Thanks! 👍

PR #431, opened 2019-01-03
valid xml

kkraune:


jobergum:
I'm uncertain about this and how it would be handled upstream in the content layer  @vekterli

PR #433, opened 2019-01-08
refactor parent/child to separate document

kkraune:
- move example in search-definitions.html to parent-child.html and build this as a separate doc. this is not yet finished, but lets merge for now, it is OK for a day or two (I will fix it quickly)
- the interesting part to review is changes in reference/search-definitions-reference.html: rewrite array, and the tables of field types


documentation/reference/search-definitions-reference.html
```diff
@@ -2644,20 +2637,29 @@ <h2 id="field_types">Field types</h2>
...
 <table class="table">
   <thead></thead><tbody>
     <tr>
       <th style="width:100px">Indexing</th>
       <td>Invalid - deployment will fail.</td>
     </tr><tr>
       <th>Attribute</th>
-      <td>As <a href="#type:string">string</a> - a reference must be an attribute.</td>
+      <td>As <a href="#type:string">string</a> - a reference must be an attribute.
+          Can be NULL or point to a non-existing document</td>
```
geirst: I think NULL can be misleading here. Consider replacing with "empty string" or "not specified"

documentation/reference/search-definitions-reference.html
```diff
@@ -2644,20 +2637,29 @@ <h2 id="field_types">Field types</h2>
...
 <tr><th id="type:reference">reference&lt;document-type&gt;</th>
 <td>
+<p>
 A <em>reference&lt;document-type&gt;</em> field is a reference to an instance of a document-type -
-i.e. a foreign key. The reference is the <a href="../documents.html">document id</a>
-of the document-type instance, hence a string.
-References are used to join documents in a
-<a href="../search-definitions.html#document-references">parent-child relationship</a>.
+i.e. a foreign key.
+The reference is the <a href="../documents.html">document id</a> of the document-type instance, hence a string attribute.
```
geirst: I think attribute should be dropped here as we are talking about the string value. The reference field is not a string attribute.

documentation/reference/search-definitions-reference.html
kkraune: agreed!

documentation/reference/search-definitions-reference.html
kkraune: you are right - corrected

PR #435, opened 2019-01-10
clean up, some links

kkraune:
search def doc should be improved, nexts PRs. parent child doc better now

### DIFF
--- a/documentation/parent-child.html
+++ b/documentation/parent-child.html
@@ -70,7 +70,7 @@ Performance notes:
 
 <h2 id="multi-value-fields">Multi-value fields</h2>
 <p>
-A document can have <a href="search-definitions.html#field">fields</a> that are arrays of structs.
+A document can have <a href="search-definitions.html#field">fields</a> that are arrays or maps of struct.
 Structs and documents are similar - a set of field name/value pairs.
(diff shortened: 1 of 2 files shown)


documentation/parent-child.html
```diff
@@ -21,84 +23,71 @@
...
+</ul>
 </p>
 
 
 
-<h2 id="multivalue-fields">Multivalue fields</h2>
+<h2 id="multi-value-fields">Multi-value fields</h2>
 <p>
-A document can have fields that are arrays of structs.
+A document can have <a href="search-definitions.html#field">fields</a> that are arrays of structs.
```
geirst: Consider adding maps of struct as well.

documentation/search-definitions.html
```diff
@@ -82,6 +69,19 @@
 
 
 
+<h2 id="field">field</h2>
```
geirst: Change to uppercase f in title?

kkraune: I could not make up my mind, as others have lowercase as it is the exact name in the search def. will revisit

documentation/parent-child.html
kkraune: yup - added

geirst:
👍

PR #437, opened 2019-01-11
Document the regexp for names

bratseth:
@arnej27959 please review


bratseth:
I don't share the belief system of those that made up that set of tags. 

And I don't think anything is deprecated.

PR #445, opened 2019-01-16
logd state port

kkraune:
- this table is most likely not complete, and it should include state ports etc for other services, too
- but not worse with this commit ...

@arnej27959 please merge


kkraune:
ping @arnej27959

arnej27959:
most state ports are dynamically allocated, but this is fixed so yes it's better to include it.

PR #448, opened 2019-01-16
Add maven dependency instead of javac + notes on shutdown MERGEOK

jobergum:


documentation/document-api-guide.html
```diff
@@ -276,6 +278,8 @@ <h3 id="asyncsession">AsyncSession</h3>
     inspecting the response's request identifier.
     The response may also contain data, either a retrieved document or a failed document put
     or update that needs to be handled.</li>
+<li>Note that the client must process the response queue or your JVM will run into garabage collection issues as the underlaying session keeps track of all responses
```
bratseth: underlying

jobergum: Fixed

PR #463, opened 2019-01-21
Add special token doc with query rewrite

kkraune:
- text-processing.html was not linked from anywhere
- this is a sister feature to query rewrite and query phrasing -> same doc

@bratseth please review the reference to isTextCharacter - not sure if this is the right definition of text in query processing.


bratseth:
> I am not sure if the fsa is auto generated when changing config? 

It's not

PR #475, opened 2019-02-04
Not using legacy name for relevance

bjormel:


bjormel:
@kkraune 
Ref 

> The document's relevance score for this query. This is the same as the default ordering when no sort specification is given ([rank] is a legacy alias for the same thing).
--

PR #476, opened 2019-02-04
Kkraune/bool int attr MERGEOK

kkraune:
@geirst the important part is to review the memory calculations, seems like same as for other numeric

@baldersheim @bratseth FYI

### DIFF
--- a/documentation/performance/attribute-memory-usage.html
+++ b/documentation/performance/attribute-memory-usage.html
@@ -23,17 +23,17 @@ The memory footprint of an attribute depends on a few factors,
 data type being the most important:
 </p>
 <ul>
-  <li>Numeric types (int, long, byte, and double) - fixed length
-      and fix cost per document</li>
+  <li>Numeric (int, long, byte, and double) and Boolean (bit) types  -
+    fixed length and fix cost per document</li>
   <li>String type - the footprint depends on the length of the
       strings and how many unique strings that needs to be
       stored.</li>
 </ul>
 <p>
 Collection types like array and weighted sets increases the memory
-usage some, but the main factor is the average number of values per
-document. String attributes are typically the largest attributes, and
-requires most memory during initialization - use numeric types where possible.
+usage some, but the main factor is the average number of values per document.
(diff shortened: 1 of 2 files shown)


documentation/performance/attribute-memory-usage.html
```diff
@@ -267,6 +259,16 @@ <h3 id="variants">Variants</h3>
       <td>Document vector, Multi-value mapping, Enum store, Posting list</td>
       <td>D * IW * ROF + D * V * EW * ROF + U * (FW+VW+4) * ROF + D * V * PW * ROF</td>
     </tr>
+    <tr>
+      <td>Boolean single-value</td>
+      <td>Document vector</td>
+      <td>D * FW * ROF</td>
+    </tr>
+    <tr>
+      <td>Boolean multi-value</td>
```
geirst: I think this still should be "Numeric multi-value value plain"?

documentation/performance/feature-tuning.html
```diff
@@ -119,4 +119,11 @@ <h2 id="document-summaries">Document summaries</h2>
 
 <h2 id="boolean-numeric-text-attribute">Boolean, numeric, text attribute</h2>
 <p>
+When selecting attribute field type, considering performance, this is a rule of thumb:
+<ol>
+  <li>Use boolean if a field is a boolean (max two values)</li>
+  <li>Use a string attribute if there is a set of values - only unique attributes are stored</li>
```
geirst: Change to "only unique strings"?

documentation/performance/attribute-memory-usage.html
kkraune: not sure what you mean? I added bool single / multi value - and removed a duplicate entry for _Numeric multi-value value plain_ at (see https://docs.vespa.ai/documentation/performance/attribute-memory-usage.html)

documentation/performance/feature-tuning.html
kkraune: willdo!

geirst:
👍

PR #481, opened 2019-02-05
vespa-visit ref doc and bucket space placeholder

kkraune:
- for later improvements for doc on visiting global docs

please comment on (some of) the Todo's - not important to cover all, this is a lot ...

content primarily from -h output - without _-e,--headersonly_ which is no more ...

### DIFF
--- a/documentation/reference/vespa-cmdline-tools.html
+++ b/documentation/reference/vespa-cmdline-tools.html
@@ -1398,7 +1398,8 @@ Options:
     </tr><tr>
       <th> -s, --bucketspace &lt;space&gt;</th>
       <td>
-          Stat buckets within the given bucket space - if not provided, 'default' is used
+          <a href="../content/buckets.html#bucket-space">Bucket space</a>
+          (<em>default</em> or <em>global</em>). If not specified, <em>default</em> is used
       </td>
     </tr><tr>
       <th> -u, --user &lt;userid&gt;</th>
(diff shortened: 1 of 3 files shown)


documentation/reference/vespa-cmdline-tools.html
```diff
@@ -1499,11 +1500,141 @@ <h2 id="vespa-status-filedistribution">vespa-status-filedistribution</h2>
...
+      </td>
+    </tr><tr>
+      <th>-c, --cluster &lt;cluster&gt;</th>
+      <td>Visit the given cluster</td>
+    </tr><tr>
+      <th>-d, --datahandler &lt;target&gt;</th>
+      <td>Send results to the given target</td> <!-- ToDo document vespa-visit-target below and link to it -->
+    </tr><tr>
+      <th>-f, --from &lt;timestamp&gt;</th>
+      <td>Only visit from the given timestamp (microseconds)</td> <!-- ToDo: Does this work now? -->
```
vekterli: As far as I know timestamp ranges should still be supported. It's effectively just a filter in the backend and therefore doesn't affect any of the distribution/id scheme stuff.

documentation/reference/vespa-cmdline-tools.html
```diff
@@ -1499,11 +1500,141 @@ <h2 id="vespa-status-filedistribution">vespa-status-filedistribution</h2>
...
+      <td>Show progress and info on STDERR</td>
+    </tr><tr>
+      <th>--visitinconsistentbuckets</th>
+      <td>Don't wait for inconsistent buckets to become consistent</td>
+    </tr><tr>
+      <th>--visitlibrary &lt;string&gt;</th>
+      <td>Use the given visitor library</td>
+    </tr><tr>
+      <th>-x, --xmloutput</th>
+      <td>Output documents as XML</td> <!-- ToDo; deprecated? -->
```
vekterli: I believe it's deprecated but since it'll likely hang around until Vespa 8 we might as well keep it documented for now (but with a deprecation comment). @bratseth is all XML rendering deprecated entirely or do we expect to keep it around as a non-default format still?

PR #493, opened 2019-02-15
Kraune/json details

kkraune:


kkraune:
@lesters not much new, I added some examples and made things consistent - so do not spend too much time reviewing the content ...

PR #494, opened 2019-02-18
match-phase max-hits is required, no default

bjormel:
Testing with unset gives 

> Error loading model: match-phase did not set max-hits > 0"


bjormel:
@baldersheim

PR #496, opened 2019-02-18
Kkraune/env vars

kkraune:


documentation/reference/services-container.html
```diff
@@ -680,14 +680,22 @@ <h2 id="nodes">nodes</h2>
 
 
 
+<h2 id="environment-variables">environment-variables</h2>
+<p>
+Add children elements to <code>nodes</code> for environment variables - see example below.
+These are set before the services are started on the container node.
+</p>
```
baldersheim: It is not available for all services. Just for the jdisc container.

kkraune:
this doc is services-container.html so implicit for container only

PR #498, opened 2019-02-19
Update an struct attribute inside a map field

fgfernandez0321:
Motivation: https://stackoverflow.com/questions/54697768/assigning-value-field-of-a-struct-in-a-map-struct-deletes-unspecified-values/54698231#54698231

It was added an example to illustrate how it can be updated an `struct` inside a `map`.

### DIFF
--- a/documentation/reference/document-json-format.html
+++ b/documentation/reference/document-json-format.html
@@ -635,7 +635,31 @@ for restrictions using structs.
 
 <h3 id="assign-map-field">Map field</h3>
 <p>
-Individual map entries are updated using <a href="document-field-path.html">field path</a> syntax.
+Individual map entries can be updated using <a href="document-field-path.html">field path</a> syntax.
+The following declaration defines a <code>map</code> where the <code>key</code> is an Integer and the value
+is a <code>person</code> struct.
+<pre>
+struct person {
+    field first_name type string {}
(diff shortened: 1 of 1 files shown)


yahoocla:

kkraune:
nice! please submit the CLA and I will merge - thanks! :-)

kkraune:
@geirst please reivew wrt attributes (not only summary)

fgfernandez0321:
I did sign the CLA using my Github account.

jobergum:
@geirst context from https://stackoverflow.com/questions/54697768/assigning-value-field-of-a-struct-in-a-map-struct-deletes-unspecified-values. What one need to watch out for here is that this particular example is a field which is summary only and the update uses fieldpath which might cause inconsistency when used with attribute. Context in https://github.com/vespa-engine/vespa/issues/8258

geirst:
@jobergum I have verified that the same field path update (`"contact{0}.first_name": { "assign": "John" }`) works on a field like this (all struct fields are defined as attributes):
`
field contact type map&lt;int, person&gt; {
      indexing: summary
      struct-field key { indexing: attribute }
      struct-field value.first_name { indexing: attribute }
      struct-field value.last_name { indexing: attribute }
    }
`

PR #504, opened 2019-02-26
Add documentation for tensor partial updates

lesters:
@geirst For discussion, please take a look. 

Don't merge until functionality is fully present.

### DIFF
--- a/documentation/reference/document-json-format.html
+++ b/documentation/reference/document-json-format.html
@@ -348,12 +348,13 @@ Manage documents using:
     <ul>
       <li><em>Assign</em>: Replace the content of the field,
         by value or by using arithmetic operations</li>
-      <li><em>Add</em>: Add a new value to a field (array, weightedset etc).
+      <li><em>Add</em>: Add a new value to a field (array, weightedset, tensor etc).
         <strong>Note:</strong> Messages can be re-sent by Vespa's Message Bus.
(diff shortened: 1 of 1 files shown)


documentation/reference/document-json-format.html
```diff
@@ -905,11 +915,40 @@ <h3 id="add-weighted-set">Add weighted set entries</h3>
 </pre>
 </p>
 
+<h3 id="tensor-add">Add tensor cells</h3>
+<p>
+Add cells to the mapped dimension of tensors. Invalid for tensors with only indexed
+dimensions. Adding a cell that already exists will overwrite the cell value with the new value.
+
+<pre>
+field tensorfield type tensor(x{},y[3]) {
```
geirst: I think we should change the type to tensor(x{},y[2]) such that the example specifies the entire indexed sub-space.

lesters: Actually, the idea here was to show default values in add, but that point was missing. Fixed now.

documentation/reference/document-json-format.html
```diff
@@ -991,7 +991,7 @@ <h4>Map field</h4>
 
 <h4 id="tensor-remove">Tensor field</h4>
 <p>
-Removes cells from the mapped dimensions of a tensor.
+Removes cells from the mapped or mixed tensors.
```
geirst: Remove 'the'?

PR #510, opened 2019-03-01
Add missing IDs used in other pages

dkurzaj:
We need to add those IDs to make the use of links to those anchors possible.

(e.g.: https://github.com/vespa-engine/documentation/blob/master/documentation/writing-to-vespa.html#L37)

Also it corresponds to the guidelines of this documentation: https://github.com/vespa-engine/documentation#links


jobergum:
Thanks!

PR #515, opened 2019-03-06
Kkraune/multivalue

kkraune:
the only interesting changes are in
* reference/search-definitions-reference.html
* search-definitions.html

there is probably things missing from the multivalue section, please comment and I will add


documentation/search-definitions.html
```diff
@@ -94,12 +94,13 @@ <h3 id="multivalue-fields">Multivalue fields</h3>
 </p><p>
 Most multivalue fields can be used in <a href="reference/grouping-syntax.html#multivalue-attributes">grouping</a>.
 </p><p>
-Multivalue fields and summaries - how to know which element hit
 </p><p>
+When searching in array of struct,
```
geirst: This also applies for maps (both of primitive types and struct).

kkraune: thanks, will add map

PR #518, opened 2019-03-08
Kkraune/minor tensor fixes

kkraune:
update note


kkraune:
@lesters FYI

kkraune:
related: the ranking doc should link to this doc - it has some tensor-specific rank functions defined here - not sure if this is linked to from the right docs?

PR #522, opened 2019-03-12
Add information about coverage object in logs

frodelu:
@kkraune / @baldersheim Please review

Re-using (pointing) to documentation already existing in https://docs.vespa.ai/documentation/graceful-degradation.html#coverage


kkraune:
bump

PR #533, opened 2019-03-27
unique id for search box input

kkraune:
this will make links like https://docs.vespa.ai/documentation/reference/search-definitions-reference.html#search work

this will not break thesearch function as the _name_ is used in search.html
```
 id="searchinput" name="q"

var query = params["q"];
```


kkraune:
ok now?

PR #535, opened 2019-03-28
Kkraune/linguistics 1

kkraune:
This is first step in revamping this doc:
* put stemming here
* OpenNlp is now default


kkraune:
merge at will, @jobergum

PR #547, opened 2019-04-05
query tracing example

kkraune:
need to add more later, also for the query explain, but this is at least something...

### DIFF
--- a/documentation/search-api.html
+++ b/documentation/search-api.html
@@ -265,3 +265,7 @@ $ curl -s 'http://host:port/search/?yql=...&amp;tracelevel=9&amp;trace.timestamp
 </pre>
 </p>
 <!-- ToDo: format=page is deprecated, so rewrite this later -->
+<p>
+Use <a href="https://javadoc.io/page/com.yahoo.vespa/container-search/latest/com/yahoo/search/Query.html">Query.trace</a>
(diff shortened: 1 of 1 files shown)


documentation/search-api.html
```diff
@@ -248,3 +250,22 @@ <h2 id="results">Results</h2>
...
 use this to write code to manipulate results.
 </p>
+
+
+
+<h2 id="query-tracing">Query tracing</h2>
+<p>
+Use query tracing to debug query execution.
+Enable by using <a href="reference/search-api-reference.html#tracelevel">tracelevel=1</a> (or higher).
+Add <a href="reference/search-api-reference.html#trace.timestamps">trace.timestamps=true</a> for timing info.
```
baldersheim: ... info for every searcher invoked.

documentation/search-api.html
```diff
@@ -248,3 +250,22 @@ <h2 id="results">Results</h2>
...
+Example:
+<pre>
+$ curl -s 'http://host:port/search/?yql=...&amp;tracelevel=9&amp;trace.timestamps=true&amp;format=page | \
```
baldersheim: Not sure if it good to use 9 as default in example. 9 gives a bit much...

PR #548, opened 2019-04-05
use json and link to result format

kkraune:
doc now much better using proper json and some text cleanup


kkraune:
ping @baldersheim

PR #549, opened 2019-04-08
Add info on term isRanked

bratseth:
@geirst


geirst:
@bratseth There are some conflicts that must be resolved

PR #551, opened 2019-04-08
Doc on counting groups 

jobergum:


jobergum:
Wait a sec, commit d3b05728e096a0ee770609b3e19be056248ec51 was not meant to be in this :)

documentation/grouping.html
```diff
@@ -1533,18 +1533,21 @@ <h3 id="time-and-date">Time and date</h3>
...
 
-Output the number of groups, which is equivalent to the number of unique values for attribute "a":
+
+Output an estimate of the number of groups, which is equivalent to the number of unique values for attribute "a":
 <pre>all( group(a) output(count()) )</pre>
-Output the number of unique string lengths for the attribute "name":
+Output an estimate of the number of unique string lengths for the attribute "name":
 <pre>all( group(strlen(name)) output(count()) )</pre>
-Output the sum of the "b" attribute for each group in addition to the overall group count:
+Output the sum of the "b" attribute for each group in addition to an estimated count of the overall number of unique groups:
```
bjorncs: It's not an estimate in this case. Since the each clause causes all groups to be returned, the estimated value is replaced by the actual group count.

jobergum: Addressed in 279d71e

PR #559, opened 2019-04-10
Fix various ranking/nativeRank documentation  MERGEOK 

jobergum:


documentation/ranking.html
```diff
@@ -509,29 +509,6 @@ <h2 id="using-query-variables">Using query variables</h2>
...
-<pre>
-field title type string {
-    indexing: index
-    rank:     literal
-}
-</pre><pre>
-0.9*fieldMatch(title) + 0.1*fieldMatch(title_literal)</pre>
-</p>
-
-
```
kkraune: so https://docs.vespa.ai/documentation/reference/search-definitions-reference.html#rank should be updated too? rank:literal is gone ?  no mention at https://docs.vespa.ai/documentation/vespa7-release-notes.html. confused

jobergum: I think since Vespa 4 we have not generated _literal fields.

jobergum: Let met get rid of the rank:literal as well

PR #571, opened 2019-04-25
Kkraune/versions

kkraune:
Only the three last commits need review

Side note: need to find out why some packages default to a vespa-6 javadoc version

As we are making changes to index format shortly, lets use this as an exercise to get this right - see my comment in last commit. The doc currently has nothing on downgrading. I think this is the best doc to put such things, open to suggestions

### DIFF
--- a/documentation/vespa-versions.html
+++ b/documentation/vespa-versions.html
@@ -7,7 +7,7 @@ title: "Vespa Versions"
 Vespa uses <a href="http://semver.org/">semantic versioning</a>.
 </p><p>
 Java APIs, web service APIs and all application package constructs are supported through a major release
-and only removed on a new release if they are already marked deprecated.
+and only removed in a new release if they are already marked deprecated.
 </p><p>
 Use of deprecated Java APIs will cause a warning on compilation,
 and use of deprecated application package constructs will cause a deprecation warning on deployment.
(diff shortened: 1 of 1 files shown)


documentation/vespa-versions.html
```diff
@@ -32,4 +32,20 @@ <h2 id="stored-data">Stored Data</h2>
 For self-hosted systems, it may be necessary to upgrade through each
 minor release rather than in larger leaps to ensure Vespa can read existing data.
 This is a good practice in any case.
+</p><p>
+When changes to the format of the stored data in Vespa are made,
+one cannot downgrade / roll back an upgrade without going through the below list
+to understand steps required:
```
bratseth: I don't see why it is helpful to ask people to go through an empty list. I suggest we spell out our contract instead: We do not guarantee that it is possible to downgrade to an earlier version of Vespa (although in most cases this is fine).

kkraune: Agree, the list was there as an example of what I think we will add in a few weeks when the bm25 feature is released - you agree we will have such a list then like "Vespa-7.x.z: change to index format blabla - make sure you do this when upgrading etc + link" (and we keep that list here)?

bratseth: Do what when upgrading do you mean? What won't work is *downgrading*, nothing special is needed to *upgrade*. I think we should just state what I write above and that's that.

kkraune: earlier, when we made changes to format, we specified that "make sure you upgrade to vespa-x.y.z before upgrading to something higher. also, make sure you cannot downgrade from x.y.z+W to x.y.z-1". Something like that (depending on the change)

I agree you can always upgrade within a minor - but you want to know that going past a specific release, downgrade is guaranteed to fail because of the change - that should be in the list

maybe this is a bit premature - I will leave this for @geirst to ponder on until we are actually making the change. we will have to signal somehow that the format is changing within releases (must keep this in mind for protocol changes, too), but not sure if this is the best place or elsewhere

PR #582, opened 2019-05-16
Deprecate summary-to and remove all traces of packet cache and cache paremeters

jobergum:
Review

### DIFF
--- a/documentation/document-summaries.html
+++ b/documentation/document-summaries.html
@@ -35,26 +35,13 @@ to generate dynamic abstracts of fields, based on search keywords.
 
 
 
-<h2 id="summary-classes-in-the-search-definition">Summary classes in the search definition</h2>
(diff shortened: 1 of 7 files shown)


jobergum:
For discussion. should we use the DEPRECATED warning or just eliminate it from doc or something else?

kkraune:
It is cleaner to keep it in doc as DEPRECATED. one day someone will be confused when things just disappear

documentation/document-summaries.html
```diff
@@ -35,26 +35,13 @@
 
 
 
-<h2 id="summary-classes-in-the-search-definition">Summary classes in the search definition</h2>
+<h2 id="summary-in-the-search-definition">Definining summary sets in the search definition</h2>
```
kkraune: id should be same as heading - easier to link to then

PR #588, opened 2019-05-23
more on multivalue fields

kkraune:
describe how to write a query to filter based on number of elements in a multivalue field - I hope this makes sense.

SEO for "rank cutoff" as well

FYI @jobergum who rewamps ranking docs these days


jobergum:
You CAN do it this way but it's not efficient do use rank-score-drop-limit as a recall operator as it just scans all matching documents.

PR #590, opened 2019-05-23
now supports endpoint argument, too

kkraune:


kkraune:
@bjormel or @jobergum please approve / merge

PR #594, opened 2019-05-31
Kkraune/visit

kkraune:


documentation/content/visiting.html
```diff
@@ -6,45 +6,84 @@
...
+<a href="../reference/document-select-language.html">document selection expression</a>.
+</p><p>
+If co-localization of documents has been used, and the selection specifies location criteria,
+the visitor client may be able to identify one or a few buckets
+that are guaranteed to include all documents that will match the expression.
+In all other cases, visiting needs to visit all buckets in the cluster.
+</p><p>
+Note that streaming search typically relies on this behavior to ensure low latency,
+but the visitor client will only be able to map visitors to a bucket subset
+if the document selection is simple.
```
vekterli: We should consider clarifying what "simple" means in practice. The document selection itself may be arbitrarily complex as long as there is an `AND` dependency on the location.

documentation/content/visiting.html
```diff
@@ -481,6 +457,43 @@ <h3 id="queueing">Queueing</h3>
...
+      <th style="white-space: nowrap">Streaming search visitor</th>
+      <td>
+        Another visitor processor is the streaming search visitor. This is run on the search containers,
+        making it transparent whether search results were created from streaming search,
+        or through indexed search in proton.
+      </td>
+    </tr><tr>
+      <th>Statistics visitor</th>
+      <td>
+        There is also a statistics visitor that just generate some statistics of stored data
```
vekterli: I'm not confident this visitor actually works these days. Do we have any system tests for it or known users? If not, we should probably remove it from the list. Especially since it's not documented how to actually use the thing.

documentation/content/visiting.html
kkraune: right. the bucket id document should be improved, so we can link to location concept there. next pr

documentation/content/visiting.html
kkraune: less is more! removed

PR #595, opened 2019-05-31
Kkraune/layout

kkraune:
@baldersheim @jobergum please merge - better looking, no real changes to content, other than using jvm XML element instead of jvmargs


kkraune:
@baldersheim and/or @jobergum ping

kkraune:
@bratseth please merge so we can get this done ...

PR #598, opened 2019-06-04
Fix page build failure for vespa stateless models

lesters:
@bratseth `{{` needs to be enclosed in {% raw %} in jekyll.


lesters:
We do this elsewhere, e.g. https://github.com/vespa-engine/documentation/blob/master/documentation/reference/tensor.html#L50-L54

PR #614, opened 2019-06-24
Kkraune/test doc

kkraune:

cleanup, remove redundancies

### DIFF
--- a/documentation/testing.html
+++ b/documentation/testing.html
@@ -11,7 +11,7 @@ Application uses the application package configuration and set up a container in
 The <code><a href="http://javadoc.io/page/com.yahoo.vespa/application/latest/com/yahoo/application/container/JDisc.html">
 JDisc</a></code> class that is accessed by
 the test through <code><a href="https://javadoc.io/page/com.yahoo.vespa/application/latest/com/yahoo/application/Application.html#getJDisc-java.lang.String-">app.getJDisc(clusterName)</a></code> -
-this class has methods for exercising all common <a href="component-types.html">component types</a>.
+this class has methods for using all common <a href="component-types.html">component types</a>.
 </p><p>
 Refer to <a href="https://github.com/vespa-engine/sample-apps/blob/master/basic-search-java/src/test/java/com/mydomain/example/MetalSearcherTest.java">
 MetalSearcherTest.java</a> for example use.
--- a/documentation/testing.html
+++ b/documentation/testing.html
@@ -208,3 +208,28 @@ public class DemoTest {
 
 }
 </pre>
+
+
+
+<h2 id="system-tests">System tests</h2>
+<p>
+A System Test is a test that tests a full Vespa application, using its endpoints.
+A System Test is invoked before a change to the production environment:
+<ol>
+  <li>Application package deployment</li>
+  <li>Vespa software upgrade</li>
+</ol>
+Vespa System Tests use <a href="https://junit.org/junit5/">JUnit 5</a> annotations.
+Tests are annotated with <code>@Test</code> as unit tests.
+The test class must be annotated with <code>@Tag("SystemTest")</code>
+to be a System Test class (and hence invoked in correct phase).
+See <code>test.categories</code> in the example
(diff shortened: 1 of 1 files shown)


documentation/testing.html
```diff
@@ -208,3 +184,28 @@ <h3 id="using-configuration-builders">Using configuration builders</h3>
...
+<p>
+A System Test is a test that tests a full Vespa application, using its endpoints.
+A System Test is invoked before a change to the production environment:
+<ol>
+  <li>Application package deployment</li>
+  <li>Vespa software upgrade</li>
+</ol>
+Vespa System Tests use <a href="https://junit.org/junit5/">JUnit 5</a> annotations.
+Tests are annotated with <code>@Test</code> as unit tests.
+The test class must be annotated with <code>@Tag("SystemTest")</code>
```
jonmv: Needs to be `ai.vespa.hosted.cd.SystemTest`, which is held in the static `ai.vespa.tenant-cd.SystemTest.name` field.

documentation/testing.html
```diff
@@ -3,59 +3,35 @@
...
-Notice the <code>com.yahoo.application.container.JDisc</code> class that is accessed by
-the test through <code>app.getJDisc(clusterName)</code>. That class has methods for
-exercising all common <a href="component-types.html">component types</a>.
+The <code><a href="http://javadoc.io/page/com.yahoo.vespa/application/latest/com/yahoo/application/Application.html">
+Application</a></code> class is useful when writing unit tests.
+Application uses the application package configuration and set up a container instance for testing.
+The <code><a href="http://javadoc.io/page/com.yahoo.vespa/application/latest/com/yahoo/application/container/JDisc.html">
+JDisc</a></code> class that is accessed by
+the test through <code><a href="https://javadoc.io/page/com.yahoo.vespa/application/latest/com/yahoo/application/Application.html#getJDisc-java.lang.String-">app.getJDisc(clusterName)</a></code> -
+this class has methods for exercising all common <a href="component-types.html">component types</a>.
```
jonmv: This sentence isn't very good English :)

documentation/testing.html
jonmv: This can be changed later, though.

documentation/testing.html
```diff
@@ -208,3 +184,28 @@ <h3 id="using-configuration-builders">Using configuration builders</h3>
...
+<ol>
+  <li>Application package deployment</li>
+  <li>Vespa software upgrade</li>
+</ol>
+Vespa System Tests use <a href="https://junit.org/junit5/">JUnit 5</a> annotations.
+Tests are annotated with <code>@Test</code> as unit tests.
+The test class must be annotated with <code>@Tag("SystemTest")</code>
+to be a System Test class (and hence invoked in correct phase).
+See <code>test.categories</code> in the example
+<a href="https://github.com/vespa-engine/sample-apps/blob/master/basic-search-java/pom.xml">
```
jonmv: This must point to the implementation we own, in `tenant-base/pom.xml`.

documentation/testing.html
```diff
@@ -208,3 +184,28 @@ <h3 id="using-configuration-builders">Using configuration builders</h3>
...
+pom.xml</a> for details. To run a system test:
+<ol>
+  <li>Deploy the application, make note of endpoints</li>
+  <li>Update <code>test-config</code> ToDo: update filename</li>
+  <li>Run tests using <pre>$ mvn test -Dtest.categories=SystemTest</pre></li>
```
jonmv: Likewise, this should be a profile from the parent pom (in `tenant-base`).

documentation/testing.html
kkraune: ok, but why ai.vespa.hosted.cd.SystemTest?  this is just instrumentation to maven for which tests to run - or is there more to it?

jonmv: Right now that's the required name (comes from the @Category, which needs a class argument). We can change it, or allow both. If it doesn't need to work today, and I guess it doesn't, then "SystemTest" is fine.

kkraune:
yes we put this out now that is not wrong for self-hosted - keeping this in one place

PR #619, opened 2019-06-26
Document comparison semantics when fields are missing

vekterli:
@kkraune please review

### DIFF
--- a/documentation/reference/document-select-language.html
+++ b/documentation/reference/document-select-language.html
@@ -107,7 +107,7 @@ addition <code>+</code> and subtractions <code>-</code> have lowest precedence.
     <td>Referencing a field that is not present in a document returns a special <code>null</code> value.
         The expression <code>music.title</code> is shorthand for <code>music.title != null</code>.
         There are potentially subtle interactions with null values when used with comparisons, see
-        <a href="#null-value-comparisons">comparisons with missing fields (null values)</a>.
+        <a href="#comparisons-with-missing-fields-null-values">comparisons with missing fields (null values)</a>.
     </td>
   </tr>
   <tr>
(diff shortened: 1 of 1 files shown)


documentation/reference/document-select-language.html
```diff
@@ -178,7 +186,45 @@ <h3 id="comparison">Comparison</h3>
 </table>
 </p>
 
-<h4>Locale / Character sets</h4>
+<h4 id="null-value-comparisons">Comparisons with missing fields (null values)</h4>
```
kkraune: id is a function of heading - in this case should be _comparisons-with-missing-fields-null-values_

documentation/reference/document-select-language.html
```diff
@@ -178,7 +186,45 @@ <h3 id="comparison">Comparison</h3>
...
+<p>
+The only comparison operators that are well-defined when one or both operands may be <code>null</code>
+(i.e. field is not present) are <code>==</code> and <code>!=</code>. Using any other comparison operators
+on a <code>null</code> value will yield a special invalid value.
+</p>
+<p>
+Invalid values may "poison" any logical expression they are part of:
+</p>
+<ul>
+  <li><code>AND</code> returns invalid iff none of its operands are false and at least one is invalid</li>
```
kkraune: iff ?

vekterli: "if and only if"

kkraune: oh ...

kkraune: ohh that is

vekterli: I see we don't use "iff" anywhere else in our documentation so I'll just use "if" instead to avoid any confusion. This isn't a formal behavior spec, so doesn't need that kind of level of detail.

documentation/reference/document-select-language.html
vekterli: Fixed

PR #620, opened 2019-06-27
Update default value to match Vespa 7 default

frodelu:
@bratseth I believe this was just an oversight when moving to Vespa 7? Text right above clearly states it is enabled by default?

FYI @kkraune


bratseth:
Yes - thanks!

PR #627, opened 2019-07-08
Add doc on that a reference field is searchable.

geirst:
@vekterli please review


geirst:
> 👍 should we mention that this is a linear scan?

Fixed

PR #642, opened 2019-08-22
Kkraune/tensor type reference

kkraune:
only the second commit needs review.

questions:
- tensorFromLabels(source, dimension) creates a tensor, I assume using the default type - we should consider adding a type argument here, too, as we cannot deduct value type?
- same goes for tensorFromWeightedSet(source, dimension). or is the tensor generated based on the weighted set / query tensor value type?

### DIFF
--- a/documentation/reference/tensor.html
+++ b/documentation/reference/tensor.html
@@ -20,6 +20,7 @@ The dimensions of a tensor defines its type, see the <a href="tensor.html#tensor
 </p>
 
 
+
 <h2 id="tensor-type-spec">Tensor type spec</h2>
 <p>
 Contained in <code><a href="search-definitions-reference.html#constant">constant</a></code>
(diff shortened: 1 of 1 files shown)


documentation/reference/tensor.html
```diff
@@ -28,41 +29,52 @@ <h2 id="tensor-type-spec">Tensor type spec</h2>
 Specifies the tensor type for a tensor.
 A tensor type contains a list of dimensions on the format:
 <pre>
-tensor(dimension-1,dimension-2,...,dimension-N)
+tensor&lt;type&gt;(dimension-1,dimension-2,...,dimension-N)
```
bratseth: I suggest "value-type" as just "type" might be confusing (as the whole string is also a "type").

documentation/reference/tensor.html
```diff
@@ -28,41 +29,52 @@ <h2 id="tensor-type-spec">Tensor type spec</h2>
...
 </pre>
+The type is one of:
```
bratseth: same: "value type"

PR #644, opened 2019-08-22
Try to fix table rendering, and mark IDs as 'required'.

gjoranv:


gjoranv:
@kkraune ping. This is an attempt to fix the rendering of the three bottom tables here: https://docs.vespa.ai/documentation/reference/services-admin.html#consumer

gjoranv:
Thanks! It fixed the rendering issue.

PR #662, opened 2019-09-18
MERGEOK Avoid using dynamic port for searchnode to get metrics. Now this will…

aressem:
… use the fixed metrics proxy port to check for active documents.


aressem:
Setting MERGE ok on this as it fixes our documentation testing and does not influence the production of release candidates on factory.

PR #672, opened 2019-09-20
Use our own container instead of an Ubuntu based one that use the Mav…

aressem:
…en Central Google mirror.

Need to watch the PR build on this one.


aressem:
This won't work. Need to find another solution.

PR #676, opened 2019-09-27
first draft attr illustration

kkraune:
I expect a couple of iterations, first cut here


geirst:
@kkraune Some comments after seeing the illustration: The width of A, B, and C should be the same. They are all memory buffers of 32-bit integers. For example B we should use array of integer (array of struct is not a native attribute type, but a collection of multiple attributes. Would be confusing to introduce that here). Also the "Attribute store" should be renamed to "Attribute multivalue mapping", and the example should contain arrays of different size (with fixed size elements inside). The enum store example is somewhat confusing as well. It indicates containing arrays, but it should be different single values of different lengths.

PR #692, opened 2019-10-10
summary store / maps

kkraune:
work in progress but good for now

### DIFF
--- a/documentation/performance/sizing-feeding.html
+++ b/documentation/performance/sizing-feeding.html
@@ -247,3 +247,52 @@ vds.idealstate.merge_bucket.done_ok
 vds.idealstate.merge_bucket.done_failed
 </pre>
 </p>
+
+
+<h3 id="summary-store">Summary store</h3>
+<p>
+When debugging update performance, it is useful to know if an update hits the
+<a href="../document-summaries.html#document-store">document store</a> or not.
+Enable <em>spam</em> log level and look for <em>SummaryAdapter::put</em> - then do an update:
+<pre>
+$ vespa-logctl searchnode:proton.server.summaryadapter spam=on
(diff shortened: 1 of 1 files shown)


documentation/performance/sizing-feeding.html
```diff
@@ -247,3 +247,52 @@ <h3 id="cluster-out-of-sync">Cluster out of sync</h3>
 vds.idealstate.merge_bucket.done_failed
 </pre>
 </p>
+
+
+<h3 id="summary-store">Summary store</h3>
```
geirst: @kkraune We should call this section "Document Store".

documentation/performance/sizing-feeding.html
```diff
@@ -247,3 +247,52 @@ <h3 id="cluster-out-of-sync">Cluster out of sync</h3>
...
+If the <a href="../document-summaries.html#document-store">document store</a>
+is accessed when evaluating the condition, performance drops.
+Conditions should be evaluated using attribute values for high performance -
+in the example above, <em>myField</em> should be an attribute.
+</p><p>
+Note: If the condition uses struct or map, values are read from the document store:
+<pre>
+    "condition": "myDoc.myMap{1} == 3"
+</pre>
+This is true even though all struct/map elements are defined as attribute.
```
geirst: Here we should replace "struct/map elements" with "struct fields", as that is how it it specified and named in the sd file.

documentation/performance/sizing-feeding.html
kkraune: agreed! wil fix

documentation/performance/sizing-feeding.html
kkraune: agreed

PR #698, opened 2019-10-16
Kkraune/content nodes ii

kkraune:
The only real change is adding doc for the get/set node/cluster state tools (based on the help text) - will reorganize this a bit, coordinate with the other docs later

### DIFF
--- a/documentation/content/admin-states.html
+++ b/documentation/content/admin-states.html
@@ -182,19 +182,17 @@ title: "Cluster and Node States"
 
 <h2 id="inspecting-and-modifying-states">Inspecting and modifying states</h2>
 <p>
-  State information is available through the
-  <a href="api-state-rest-api.html">State Rest API</a>. This interface can
-  also be used to set user states for nodes.
-  For convenience, use the
-  command line tools. See the <em>--help</em> pages for
-  <em>vespa-get-cluster-state</em>, <em>vespa-get-node-state</em>
-  and <em>vespa-set-node-state</em>.
+State information is available through the
+<a href="api-state-rest-api.html">State Rest API</a>.
+This interface can also be used to set user states for nodes.
+Alternatively, use the tools
+<a href="../reference/vespa-cmdline-tools.html#vespa-get-cluster-state">vespa-get-cluster-state</a>,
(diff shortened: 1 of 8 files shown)


documentation/content/admin-states.html
```diff
@@ -4,197 +4,202 @@
...
-  Node states are input to the algorithm.
-  Distributors use the distribution algorithm to keep bucket copies on correct
-  storage nodes. To do this, it needs to know which storage nodes are available
-  to hold copies and what their capacities are.
-  Service layer nodes may split requests between different partitions, in which
-  case it needs to know what partitions are available.
-  To achieve this, state is tracked for all the major components of the content layer:
+Clients route requests to distributors using the distribution algorithm.
+Node states are input to the algorithm.
+Distributors use the distribution algorithm to keep bucket copies on correct storage nodes.
```
vekterli: As a follow-up for later, for historical reasons we are currently using (at least) 4 terms for content nodes somewhat interchangeably in our documentation:
1) storage nodes
2) service layer nodes
3) content nodes
4) search nodes

We should find one and stick to it

documentation/operations/admin-procedures.html
```diff
@@ -20,8 +20,8 @@ <h2 id="system-status">System status</h2>
   <li>Use <a href="../reference/search-api-reference.html#tracelevel">query tracing</a></li>
   <li>Use <a href="../reference/services.html#clients">feed tracing</a></li>
     <!-- ToDo: this works? must document ... -->
-  <li>Use the <a href="../elastic-vespa.html#cluster-controller">
-    Cluster Controller</a> to track the status of search/storage nodes.</li>
+  <li>Use the <em>cluster controller</em> status page (below)
```
vekterli: An explicit link would probably be better than just "(below)" since it lets us scan for links to replace if the content moves again

documentation/content/admin-states.html
kkraune: very much agree, sir. imho, _content_ node is what is configured in services.xml, and we can use proton in _search_ context. So for me, number 3 is the winner, but open for discusssion

documentation/operations/admin-procedures.html
kkraune: ok, considered for next pr

PR #705, opened 2019-10-23
Kkraune/node states

kkraune:
some questions in the distributor section

### DIFF
--- a/documentation/content/content-nodes.html
+++ b/documentation/content/content-nodes.html
@@ -184,6 +184,10 @@ they get the cluster state from the distributors - <a href="#distributor">detail
   e.g, distributors and content nodes that are on the same node often stop at the same time.
   </p><p>
   The version number is upped, and the new cluster state is broadcast.
+  </p><p>
+  If cluster state version is <a href="../operations/admin-procedures.html#cluster-state">reset</a>,
+  clients to distributors can temporarily fail operations in the transition,
+  but will eventually converge on the new (lower) cluster state version.
   </p>
   </td>
 </tr>
@@ -198,9 +202,9 @@ Reads and writes will work well in case of cluster controller down,
 but other changes to the cluster (like a content node down) will not be handled.
 It is hence recommended to configure a set of cluster controllers.
 </p><p>
-A set of cluster controllers elect a master among themselves.
-The master does the node polling and cluster state broadcast.
-The other cluster controller instances exist to participate in master election
+The cluster controller nodes elect a master,
+which does the node polling and cluster state broadcast.
+The other cluster controller nodes only exist to do master election
 and potentially take over if the master dies.
 </p><p>
 All cluster controllers will vote for cluster controller with the lowest index that says it is ready.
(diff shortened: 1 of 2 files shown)


documentation/content/content-nodes.html
```diff
@@ -213,94 +217,69 @@ <h3 id="master-election">Master election</h3>
...
+a distributor will then handle a new set of buckets.
 </p><p>
-The distribution algorithm also calculates what content nodes copies of buckets should be stored on.
-As moving bucket copies take time, the distributors track current state in a bucket database,
-allowing requests to be processed independent of where buckets are currently located.
-Distributors use the distribution algorithm to detect buckets that are stored on wrong nodes,
-and move copies to correct nodes when there are free resources to do so.
+Document API clients do not interface with the cluster controller,
+and does hence not know the cluster start at startup.
+A random distributor is used first. <!-- ToDo: Or does it use the ideal state algo? -->
```
vekterli: When a client starts up it does not have the cluster state information required to use the ideal state algorithm, so it has no choice other than to send to random node in the config (which it does have).

documentation/content/content-nodes.html
```diff
@@ -213,94 +217,69 @@ <h3 id="master-election">Master election</h3>
...
-During this reshuffling, clients may get some requests temporarily failing
-as distributors taking over bucket ownership doesn't yet know where copies of its buckets are located.
-These requests will automatically be resent, providing timeouts allow for it.
-Changes in what distributors are up and available will currently cause windows of availability loss.
-Other distributors take over bucket ownership.
-To do this, they fetch bucket metadata for new buckets from all storage nodes,
-in-memory operations for speed.
+Distributors do not persist the bucket database,
+the bucket-to-content-node mapping is stored in memory in the distributor.
+Document count per bucket is stored as well. <!-- ToDo: checksum? -->
```
vekterli: Currently document count, persisted size and a metadata checksum is stored per replica per bucket

documentation/content/content-nodes.html
```diff
@@ -213,94 +217,69 @@ <h3 id="master-election">Master election</h3>
...
+    <li>If buckets have too few replicas, new are generated on other content nodes.</li>
+  <li>If the replicas differ, a bucket merge is issued to get replicas consistent.
+      The distributor sets a bucket replica <em>trusted</em>. <!-- ToDo: need something here - this is confusing - old text below:
+      
+      For instance it may know that of the currently existing replicas,
+      a given replica has been available all the time and may be trusted to have all the information,
+      while another replica may be on a content node that recently restarted,
+      so that replicas may lack some documents.
+      Such historic state is not persisted,
+      and is thus lost on distributor restarts.--></li>
```
vekterli: The "trusted" replica notion is iffy as it's a transient property and therefore is unreliable (ironically enough, given the name) during node up/down edges. I suggest we remove the sentence mentioning it entirely.

documentation/content/content-nodes.html
```diff
@@ -213,94 +217,69 @@ <h3 id="master-election">Master election</h3>
...
-    the buckets are split or joined to remove such overlapping buckets</li>
-  <li>If buckets contain too little or too much data, they should be joined or split</li>
-  <li>If not exactly one copy is marked active, and the backend wants a copy to
-    be marked active, activate or deactivate copies to get in a good state</li>
+      the buckets are split or joined to remove such overlapping buckets.
+      Read more on <a href="buckets.html">inconsistent buckets</a>.</li> <!-- ToDo: change this to a proper anchor within buckets.html - this needs a dedicated section, relevant to point to from visiting -->
+  <li>If buckets are too small/large, they will be joined or split.</li>
+  <li>If not exactly one replica is marked active, <!-- ToDo: what is this ?-->
+      and the content node wants a replica to be marked active,
+      activate or deactivate replicas to get in a good state.</li>
```
vekterli: In clusters a with document type marked as `mode="index"` the distributor will ensure that only one of a bucket's replicas within a group will be used when processing search queries. Otherwise search results would contain many duplicates. This is known as an "active" replica, the rest being "inactive" and available for failover if the currently active node should become unavailable, or the ideal state changes.

kkraune: of course - it is that _active_ - will link to the proton doc for details here. thanks!

PR #706, opened 2019-10-23
Kkraune/buckets i

kkraune:
cleanup only - best reviewed per commit - makes next PR easier

### DIFF
--- a/documentation/content/buckets.html
+++ b/documentation/content/buckets.html
@@ -4,12 +4,12 @@ title: "Buckets"
 ---
 
 <p>
-The content layer splits the document space into chunks called <em>buckets</em>.
-</p><p>
-Documents are mapped to buckets.
+The content layer splits the document space into chunks called <em>buckets</em>,
+documents are hence mapped to buckets.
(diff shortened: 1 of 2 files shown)


documentation/content/buckets.html
```diff
@@ -5,31 +5,9 @@
...
-<ul>
-  <li>Per unit metadata takes up less space,
-    simplifying caching and metadata transfers.</li>
-  <li>Iterating metadata for all units is faster.</li>
-</ul>
-To strike a balance, documents are mapped to a set of buckets,
-where the number of buckets can be configured.
-As clusters grow or shrink, or document sizes in the cluster change,
+Documents are mapped to buckets.
+As clusters grow or shrink, or document counts change,
```
vekterli: Could consider emphasising that all this neat stuff happens without any required intervention. E.g.:
> Documents are mapped to buckets. The cluster automatically splits and joins buckets to maintain an even distribution across all nodes and to keep bucket sizes within configurable limits.

kkraune: agreed!

PR #711, opened 2019-11-04
first cut batch delete

kkraune:


kkraune:
was not sure either. considered in the document/v1, but not optimal there, too

kkraune:
https://docs.vespa.ai/documentation/performance/sizing-feeding.html maybe? I don't want to start a new doc just for this

jobergum:
Closest there is without forking a new doc.

PR #716, opened 2019-11-07
Write what happens with searchable-copies and redundancy with groups

oyving:


documentation/reference/services-content.html
```diff
@@ -387,6 +387,8 @@ <h2 id="group">group</h2>
 Can not be used in conjunction with the <code><a href="#nodes">nodes</a></code> element.
 If a non-flat structure is desired, use this element instead.
 Groups can contain other groups or nodes, but not both.
+When using groups, <code><a href="#searchable-copies">searchable-copies</a></code> and
```
bratseth: There are two ways to configure groups, one for cloud and one for self-hosted. In the cloud case, redundancy, final redundancy and searchable-copies is per group, while in the self-hosted case it's across all groups. Perhaps we need to say something about both here now.

oyving: Something like the change below?

kkraune: you can alternatively add to https://cloud.vespa.ai/reference/services-hosted.html and link to this from the open source doc

bratseth: I think it might be confusing to make the distinction be between "open source", since it's the same open source code base in both cases .... we're going to have a thousand of these so we better get it right. I suggest "when self-hosting"/"in self-hosted systems" vs. "cloud"/"Vespa cloud".

documentation/reference/services-content.html
```diff
@@ -448,53 +451,15 @@ <h2 id="distribution">distribution (in group)</h2>
...
-If the redundancy at some level is lower than the partition spec,
-Vespa stores less copies than the partition spec - starting leftwards.
-If the spec is <em>2|2|*</em> and configures to store 3 copies only,
-Vespa first fills the primary group with 2,
-then gives the secondary group the 1 remaining copy,
-and there will be no third group as there are no more copies to store.
-This may be a valid case if using multiple levels of groups
-and a single group needs to store differently for buckets depending on
-which buckets it is primary or secondary group for.
+The only supported use for the partition specification is to have as many
```
bratseth: When you are removing documentation for deeper hierarchies it's not necessary to imply that there are other uses that are not supported. Just say "The partition specification is used to ..." or something like that.

PR #725, opened 2019-11-18
more use cases

kkraune:


kkraune:
yes, not strictly use cases - these are more operational so let me move that to another doc on cloud.vespa.ai

kkraune:
PTAL

PR #731, opened 2019-11-22
simplify max-length doc

kkraune:
in #729  we removed _ fieldmatchmaxlength_ documentation - remove here too. not sure if the _max-length_ needs more text, this removes the broken link


kkraune:
or @bjormel

kkraune:
probably. I just did this to silence the link checker. can add more info later

PR #739, opened 2019-12-16
Clarify TLS setup

tokle:


tokle:
there are cases where a keystore / truststore may contain expired certs.

PR #763, opened 2020-01-07
fix for the XSS vulnerability

StevenRowe:
The search function is vulnerable for XSS attacks. This is going to fix this issue.


kkraune:
Thanks a lot, @StevenRowe  !

PR #769, opened 2020-01-09
Kkraune/ranking intro

kkraune:
the intro is new, the rest is just reordering

fyi @lesters

### DIFF
--- a/documentation/ranking.html
+++ b/documentation/ranking.html
@@ -167,6 +167,69 @@ This will make the explicitly listed rank features the only ones dumped when req
 </p>
 
 
+<h3 id="normalization">Normalization</h3>
+<p>
+The rank features provided includes both features normalized to the range 0-1,
+and un-normalized features like counts and positions.
+Whenever possible, prefer the normalized features.
+They capture the same information (and more),
+but are easier to use because they can be combined more easily with other features.
+In addition, try to write ranking expressions such that the combined rank score is also normalized,
+for example by taking averages not sums.
+The resulting normalized rank scores makes it possible to implement relevance based blending,
+search assistance triggering when there are no good hits, and so on.
+</p>
+
+
+<h3 id="configuration">Configuration</h3>
+<p>
+Some features, most notably the <a href="text-ranking.html">fieldMatch</a> features,
+has configuration parameters that enables the feature calculation to be tweaked per field for performance or relevance.
+Feature configuration values are set by adding to the rank profile:
+<pre>
+rank-properties {
+    featureName.configurationProperty: "value"
+}
+</pre>
+The values are set per field, like:
+<pre>
+rank-properties {
+    fieldMatch(title).maxAlternativeSegmentations: 10
+    fieldmatch(title).maxOccurrences: 5
+    fieldMatch(description).maxOccurrences: 20
+}
+</pre>
+Refer to the
+<a href="reference/rank-feature-configuration.html">rank feature configuration</a> reference.
+</p>
+
+
+<h3 id="feature-contribution-functions">Feature contribution functions</h3>
+<p>
+Vespa ranking features are linear.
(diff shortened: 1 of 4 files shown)


documentation/ranking.html
```diff
@@ -4,27 +4,91 @@
...
-An application can have multiple rank profiles -
-this can be used for implementing different use cases, or bucket testing ranking variations.
-</p><p>
-Example: Rank the freshest document highest, using the value in an attribute named <em>timestamp</em>:
+Vespa defines Big Data Serving as:
+<blockquote style="background: #1a7db6; color: #ffffff"><strong>
+Selection, organization and machine-learned model inference,
+<ul>
+    <li>over many, constantly changing data items (thousands to billions),</li>
+    <li>with low latency (~100 ms) and high load (thousands of queries/second)</li>
```
jobergum: Not sure if we should mention 100 ms explicit, it's quite high for thousands of queries per second.

documentation/ranking.html
```diff
@@ -4,27 +4,91 @@
...
+        second-phase {
+            expression: sum(onnx("my-model.onnx", "add"))
+        }
+    }
+}
+</pre>
+Applications use the <a href="search-api.html">query API</a> to <em>select</em> the documents to evaluate
+using a <a href="query-language.html">query language</a>,
+and choose a <a href="#rank-profile">rank profile</a> for the ranking.
+Rank profiles have <a href="#two-phase-ranking">one or two phases</a>:
```
jobergum: This could indicate that you must have a phased rank profile, you don't. Also suggest to just point to phased ranking https://docs.vespa.ai/documentation/phased-ranking.html

documentation/ranking.html
```diff
@@ -4,27 +4,91 @@
...
+<ul>
+    <li>Phase one should use a computationally inexpensive function to rank candidates</li>
+    <li>Phase two is run on a small candidate set</li>
+</ul>
+Hence, the query selection and first phase ranking reduces the size of the computation -
+then machine-learned models can be used on the second-phase ranking on
+<a href="reference/search-definitions-reference.html#rerank-count">rerank-count</a> documents per node.
+This makes the ranking scalable:
```
jobergum: Just point to sizing doc. Also phased ranking is about fine tuning the precision (most relevant first) while first-phase is about recall (finding the relevant docs)

documentation/ranking.html
```diff
@@ -4,27 +4,91 @@
...
+<ul>
+    <li>Control the second phase candidate set size</li>
+    <li>Add content nodes to rank more documents / less documents per node</li>
```
jobergum: More nodes => rank less per node/cpu thread if in use with multiple threads per search

documentation/ranking.html
```diff
@@ -4,27 +4,91 @@
...
+</p><p>
+Note: Vespa also supports <a href="stateless-model-evaluation.html">stateless model evaluation</a> -
+making inferences without documents (i.e. query to model).
+</p>
+
+
+
+<h2 id="machine-learned-model-inference">Machine-Learned model inference</h2>
+<p>
+Vespa supports the following ML models:
```
jobergum: Not exclusive list, we can represent lots of model, logistic regression etc. Also mention PyTorch through ONNX.

documentation/ranking.html
kkraune: Agree. But this phrase is used many places elsewhere in presentations / documents. Lets fine tune in next PR

documentation/ranking.html
kkraune: removed redundant phased ranking text and linked instead. changed the intro a little, not 100% happy with it, will iterate

documentation/ranking.html
kkraune: added todo

PR #771, opened 2020-01-10
change #3F9DD8 to --vespa-color-light-blue: #1a7db6

kkraune:
1: we should use variables for simpler management of colors
2: try out the new style guide


ldalves:
Nice. :) 

- Following the _first_ recommendation from the Sunnyvale design team, this would be still correct... 
- Following the _new_ recommendation we would need _also_ to change the background color or the logo color of the topbar in order to be correct....

PR #781, opened 2020-01-13
temporarily revert to previous navbar color

kkraune:
to not violate blue logo on dark color restriction

but I really do not like this way either (yet another blue, which is not in the updated guide), so lets think of alternatives - maybe white logo or something else


ldalves:
Agree.. i think if we have a guide, we should follow it... we just need to decide what we want... :)

PR #797, opened 2020-01-16
Fix closing tag

IcaroJerry:


documentation/performance/sizing-examples.html
```diff
@@ -166,6 +168,8 @@ <h3>9 nodes, 9 groups with 1 node per group</h3>
       &lt;node hostalias="container1"/&gt;
       &lt;node hostalias="container2"/&gt;
     &lt;/nodes&gt;
+  &lt;/container&gt;
```
bratseth: This one should be "content"

IcaroJerry: Are you sure?
I think the two "container" tags are not closing.
Look:
image

PR #801, opened 2020-01-20
Kkraune/remove redundant group setup doc

kkraune:
the new sizing examples are much better, kill the other confusing doc!


jobergum:
Great 👍 🥇

PR #802, opened 2020-01-20
Kkraune/feeding

kkraune:
more details on document expiry and docs that are silently dropped due to selection

the changes in reference/services-content.html are for the most visual, less grim now. improved the parts on selection string there


jobergum:
👍

PR #804, opened 2020-01-21
Kkraune/feed sizing

kkraune:
first cut is rewriting the intro and a more basic approach to understanding how writes are made

the work is not done, must complete in later PRs

this PR also cleans up some of the pages linked to

I will have @geirst and/or @baldersheim review when done - they have verified the changes done so far, I discussed with them before writing


jobergum:
👍

PR #817, opened 2020-01-24
move reference doc to reference. more referency format

kkraune:
no content changes, other than using the name _/document/v1/ more to disambiguate from Document API in java


bratseth:
👍 👍

PR #818, opened 2020-01-24
Improve grouping documentation

bratseth:


documentation/reference/select-reference.md
```diff
@@ -96,6 +102,24 @@ A grouping, that will group first by year and then by month, can be written as s
...
+all(group(predefined(foo, bucket[1, 2>, bucket[3, 4>)))
+```
+equivalent JSON `grouping`-argument:
+
+```
+"grouping" : [ 
+  { 
+    "all" : { 
+      "group" : { 
+        "predefined" : [ "foo", { "bucket": [1,2]}, { "bucket": [3,4]} ]
```
arnej27959: hmm... how do you differentiate "bucket<1,2>" versus "bucket[1,2>" versus "bucket[1,2]" in json?

bratseth: You are right one still do not have quite the same level of expressive power here - we'd need to represent that as additional properties instead. Anyway, this is an improvement and we're getting into quite sparsely travelled territory ...

bratseth:
PTAL @arnej27959

PR #823, opened 2020-01-27
typo

kkraune:
- probably a cut and paste error

### DIFF
--- a/documentation/reference/search-definitions-reference.html
+++ b/documentation/reference/search-definitions-reference.html
@@ -2604,8 +2604,8 @@ struct person {
 }
 field identities type map&lt;string, person&gt; {
     indexing: summary
-    struct-field key { indexing: attribute }
-    struct-field value.last_name { indexing: attribute }
+    struct-field key   { indexing: attribute }
+    struct-field value { indexing: attribute }
 }
 </pre>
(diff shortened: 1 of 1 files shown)


documentation/reference/search-definitions-reference.html
```diff
@@ -2604,8 +2604,8 @@ <h2 id="field-types">Field types</h2>
 }
 field identities type map&lt;string, person&gt; {
     indexing: summary
-    struct-field key { indexing: attribute }
-    struct-field value.last_name { indexing: attribute }
```
geirst: This is not a typo. See https://github.com/vespa-engine/system-test/blob/master/tests/search/struct_and_map_types/attribute_fields/test.sd#L32 for similar example.

geirst:
> It looks right _as it was_ to me. @geirst ?

Yes, you are right @bratseth

documentation/reference/search-definitions-reference.html
kkraune: you are of course right. the problem with the examples is the inverted order - it should start with primitive types first, and then add complications like structs (I confused myself reading this). @jobergum har asked for a separate guide for multivalue fields - I think we could stuff searchdef examples there and link this this from the reference instead

PR #826, opened 2020-01-29
link to javadoc, use examples - remove other text

kkraune:
I promise to put samples in a sampleapp instead - paste here for now while I develop it


kkraune:
Thanks! Good point - https://docs.vespa.ai/documentation/writing-to-vespa.html says _Note that only assign and remove are idempotent - message re-sending can apply updates more than once. Use conditional writes for stronger consistency._

PR #837, opened 2020-02-06
Add doc for imported fields in document selections

vekterli:
@kkraune @geirst please review

### DIFF
--- a/documentation/documents.html
+++ b/documentation/documents.html
@@ -217,13 +217,22 @@ Notes:
     and have a major performance impact on the system.
   </li><li>
     <a href="reference/search-definitions-reference.html#import-field">Imported fields</a>
-    can not be used in the selection string to expire documents
+    can be used in the selection string to expire documents, but special care needs to be
+    taken when using these.
+    See <a href="reference/document-select-language.html#using-imported-fields-in-selections">
+    using imported fields in selections</a> for more information and restrictions.
   </li>
 </ul>
 To batch remove, set a selection that matches no documents, like <em>"!music"</em>
 </p><p>
-Use <a href="content/visiting.html">vespa-visit</a> to test the selection (dump all IDs):
+Use <a href="content/visiting.html">vespa-visit</a> to test the selection. Dump the IDs of all documents that would be <em>preserved</em>:
+</p>
 <pre>
 $ vespa-visit -i -p progress-file -s 'music.timestamp &gt; now() - 86400' > ids.json
 </pre>
+<p>
+Negate the expression by wrapping it in a <code>not</code> to dump the IDs of all the documents that would be <em>removed</em> during GC:
 </p>
+<pre>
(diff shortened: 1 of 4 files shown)


documentation/documents.html
```diff
@@ -217,13 +217,22 @@ <h2 id="document-expiry">Document expiry</h2>
...
 To batch remove, set a selection that matches no documents, like <em>"!music"</em>
 </p><p>
-Use <a href="content/visiting.html">vespa-visit</a> to test the selection (dump all IDs):
+Use <a href="content/visiting.html">vespa-visit</a> to test the selection. Dump the IDs of all documents that would be <em>preserved</em>:
+</p>
 <pre>
 $ vespa-visit -i -p progress-file -s 'music.timestamp &gt; now() - 86400' > ids.json
 </pre>
+<p>
+Negate the expression by wrapping it in a <code>not</code> to dump the IDs of all the documents that would be <em>removed</em> during GC:
```
kkraune: great! makes it clearer, and how to test, too :-)

documentation/reference/document-select-language.html
```diff
@@ -389,7 +409,79 @@ <h3 id="now-function">Now function</h3>
...
+<p><code>child.a1 &gt; 5</code></p>
+<p>Find all child docs whose parents have an <code>a2</code> of 10 and grandparents have <code>a1</code> of 4:</p>
+<p><code>child.a1 == 10 and child.a2 == 4</code></p>
+<p>Find all child docs where the parent document cannot be found (or where the referenced field is not set in the parent):</p>
+<p><code>child.a2 == null</code></p>
+<p>Note that when visiting <code>child</code> documents we only ever access imported fields via the
+<strong>child</strong> document type itself.</p>
+<p>A much more complete list usage examples for the above document schemas and
+reference relations can be found in the
+<a href="https://github.com/vespa-engine/system-test/blob/master/tests/search/parent_child/imported_fields_in_selections.rb">
```
kkraune: good idea to link to test!

documentation/reference/document-select-language.html
```diff
@@ -223,6 +236,13 @@ <h4 id="comparisons-with-missing-fields-null-values">Comparisons with missing fi
 To work around this issue, "short-circuiting" using a field presence check may be used:
 <code>foo.expires_at_time and foo.expires_at_time > now()</code>.
 </p>
+<div class="alert alert-warning">
+If your selection references imported fields, <code>null</code> will be returned for any imported field when
```
geirst: Perhaps we should add a short example on how such GC expression should be formulated (taking null into account)? Could even use the grandparent / parent / child setup described further down.

vekterli: There is an example in the paragraph above it, but it would make sense to add one explicitly

vekterli:
@kkraune @geirst I've explicitly added examples of `null` checking in GC expressions, PTAL

PR #849, opened 2020-02-17
indexing docproc

kkraune:


bjormel:
Better, while you are at it, would it make sense to add some information about processing here too? https://docs.vespa.ai/documentation/writing-to-vespa.html

kkraune:
good point, @bjormel , PTAL

PR #852, opened 2020-02-18
count all docs example

kkraune:


bjormel:
Should we add 'limit 0', to avoid docsum fetching for a total count request?

kkraune:
it just complicates for the simple example ... this is manual stuff normally

kkraune:
but technically you are right. we should maybe mention this elsewhere

bjormel:
👍

PR #887, opened 2020-03-09
Introduce new 'monitoring' document.

gjoranv:
I will create PRs to update the links in our blog posts to the moved sections.

FYI: @oyving

### DIFF
--- a/documentation/monitoring.html
+++ b/documentation/monitoring.html
@@ -1,0 +1,108 @@
+---
+# Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+title: "Monitoring"
+---
+
+<p>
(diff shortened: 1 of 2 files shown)


documentation/monitoring.html
```diff
@@ -0,0 +1,108 @@
...
+<h3 id="pushing-to-cloudwatch">Pushing metrics to CloudWatch</h3>
+<p class="alert alert-success">
+    This method currently works for self-hosted Vespa only.
+</p>
+<p>
+    This is presumably the most convenient way to monitor Vespa in CloudWatch. You will need to have the following
+    available:
+</p>
+<ul>
+    <li>An IAM user that only has the <em>putMetricData</em> permission.</li>
```
olaaun: They could also use an IAM role as long as they have a mechanism to keep the credentials file updated.

PR #900, opened 2020-03-15
note on EBS and Burst Balance

kkraune:


jobergum:
I think this belongs in a hw-flavor document but ok :)

kkraune:
> I think this belongs in a hw-flavor document

agreed. Once we have added this in a doc , we can move this and link to it instead

PR #951, opened 2020-04-21
Remove (partially) incorrect normalization table

kkraune:
- no need for it - link to implementation instead, that again links to UNICODE NORMALIZATION FORMS

FYI @geirst

### DIFF
--- a/documentation/linguistics.html
+++ b/documentation/linguistics.html
@@ -210,7 +210,7 @@ use <code><a href="reference/schema-reference.html#indexing-rewrite">indexing-re
 </p><p>
 Also see <a href="reference/schema-reference.html#gram">N-gram matching</a>.
 </p><p>
-Tokenization is not supported in <a href="streaming-search.html">streaming search</a>.
+Tokenization is not used in <a href="streaming-search.html">streaming search</a>.
 </p>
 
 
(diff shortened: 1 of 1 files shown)


documentation/linguistics.html
```diff
@@ -205,89 +205,33 @@ <h2 id="tokenization">Tokenization</h2>
...
 The resulting tokens are then searchable in the index.
 </p><p>
-Not supported in <a href="streaming-search.html">streaming search</a>.
-</p><p>
 To index strings as-is (that is, avoid tokenization),
 use <code><a href="reference/schema-reference.html#indexing-rewrite">indexing-rewrite</a>: none</code>.
 </p><p>
 Also see <a href="reference/schema-reference.html#gram">N-gram matching</a>.
+</p><p>
+Tokenization is not supported in <a href="streaming-search.html">streaming search</a>.
```
bratseth, line 213: I think we should say "not used" here.

documentation/linguistics.html
```diff
@@ -205,89 +205,33 @@ <h2 id="tokenization">Tokenization</h2>
...
-        </tr>
-        <tr>
-            <td></td>
-            <td>ï &#x21D2; i</td>
-            <td></td>
-            <td></td>
-        </tr>
-    </tbody>
-</table>
+Normalization is not supported in <a href="streaming-search.html">streaming search</a>.
```
bratseth, line 230: Same here; "not used".

documentation/linguistics.html
kkraune: yes, better

PR #953, opened 2020-04-21
prefix details and some cleanup

kkraune:

### DIFF
--- a/documentation/text-matching-ranking.html
+++ b/documentation/text-matching-ranking.html
@@ -138,14 +138,15 @@ Also try query tracing to see how query parsing changes with
 <h3 id="prefix-search">Prefix Search</h3>
 <p>
 A prefix search will match the query term "hea" to documents with "head".
-However, prefix search is <span style="text-decoration: underline"> not</span>
-enabled in text matching using <em>index</em> mode - some alternatives:
+However, prefix search is <span style="text-decoration: underline">not</span>
+supported in text matching using <em>index</em> mode - some alternatives:
 </p>
 <table class="table">
   <tr>
     <th style="white-space: nowrap">Array of string attributes</th>
     <td>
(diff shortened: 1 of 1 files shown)


documentation/text-matching-ranking.html
```diff
@@ -134,12 +135,77 @@ <h3 id="query-trace">Query Trace</h3>
...
-<!-- ToDo: streaming search note -->
-<!-- ToDo: refer to
-  https://docs.vespa.ai/documentation/reference/schema-reference.html#match
-  https://docs.vespa.ai/documentation/reference/schema-reference.html#indexing
-  and write some guidance -->
+<h3 id="prefix-search">Prefix Search</h3>
+<p>
+A prefix search will match the query term "hea" to documents with "head".
+However, prefix search is <span style="text-decoration: underline"> not</span>
+enabled in text matching using <em>index</em> mode - some alternatives:
```
geirst, line 142: Consider replacing "enabled" with "supported" as someone might interpret this as it can be turned on.

documentation/text-matching-ranking.html
```diff
@@ -134,12 +135,77 @@ <h3 id="query-trace">Query Trace</h3>
...
+</p>
+<table class="table">
+  <tr>
+    <th style="white-space: nowrap">Array of string attributes</th>
+    <td>
+      Use an attribute with array of strings in addition to the index string field.
```
geirst, line 148: Consider mention that prefix search is also supported for single string attributes.

documentation/text-matching-ranking.html
kkraune: yup

documentation/text-matching-ranking.html
kkraune: will fix

PR #960, opened 2020-04-24
improve streaming search doc

kkraune:
thanks @geirst for helping with this!

### DIFF
--- a/documentation/streaming-search.html
+++ b/documentation/streaming-search.html
@@ -25,7 +25,7 @@ Only data structures for document IDs are in memory,
 not <a href="attributes.html">attributes</a>.
 It matches documents to queries by streaming through them, similar to a grep.
 This is too costly for a global search but works fine for searching small subsets of the
-data. This means Vespa can avoid the overhead of maintaining reverse indexes.
+data. This means Vespa can avoid the overhead of maintaining indexes.
 Streaming mode is suitable when subsets are <em>on average</em> very small compared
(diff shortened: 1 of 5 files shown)


kkraune:
ping. @geirst  feel free to merge, too :-)

documentation/streaming-search.html
```diff
@@ -82,29 +46,112 @@
...
+Next step is setting correct <a href="reference/schema-reference.html#match">match mode</a> for fields -
+example using default string tokenized matching:
+<pre>
+field artist type string {
+    indexing: summary | index
+}
+
+# Use <a href="reference/vespa-cmdline-tools.html#vespa-configproxy-cmd">vespa-configproxy-cmd</a> to find the value for the -i argument below
+
+$ <a href="reference/vespa-cmdline-tools.html#vespa-get-config">vespa-get-config</a> -n vespa.config.search.vsm.vsmfields -i music/search/cluster.music.music | egrep 'name|arg1'
```
geirst, line 108: I am not sure we want to put this information here. This is for debugging and might confuse the reader. What is relevant is what we put in "match:" in the field definition. The same applies for the examples further down. Perhaps add a debug section at the bottom with one example using vespa-get-config.

kkraune: good point! that makes it more readable. I included it here to make it easier for people to see actual settings. will change in next PR

PR #966, opened 2020-04-29
Kkraune/match debug

kkraune:
FYI @jobergum 

new section for what to do when debugging matching. the other changes are reordering / moving text + some cleanup

### DIFF
--- a/documentation/text-matching-ranking.html
+++ b/documentation/text-matching-ranking.html
@@ -121,6 +121,9 @@ Read more about the attribute <a href="reference/schema-reference.html#match">wo
 </p>
 
 
+<!-- ToDo: Add more match examples for indexed search-->
+
+
 <h3 id="query-trace">Query Trace</h3>
 <p>
 Adding <a href="reference/query-api-reference.html#tracelevel">tracelevel=2</a>
--- a/documentation/text-matching-ranking.html
+++ b/documentation/text-matching-ranking.html
@@ -207,6 +207,69 @@ ull	1
 
 
 
+<h2 id="match-configuration">Match Configuration</h2>
+<p>
+Indexed and streaming search differ in execution
+and the match-configuration differs.
+Some times it is useful to inspect generated configuration
+to understand or validate the match configuration.
+Run this to find the value of the -i argument below:
+<pre>
+$ <a href="reference/vespa-cmdline-tools.html#vespa-configproxy-cmd">vespa-configproxy-cmd</a>
+</pre>
+</p>
+
+
+<h3 id="indexed-search">Indexed Search</h3>
+<p>
+Note the difference for the <em>artist</em> and <em>album</em> fields:
+<pre>
+field artist type string {
+    indexing: summary | index
+    match   : exact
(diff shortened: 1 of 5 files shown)


documentation/text-matching-ranking.html
```diff
@@ -207,6 +207,69 @@ <h3 id="prefix-search">Prefix Search</h3>
 
 
 
+<h2 id="match-configuration">Match Configuration</h2>
+<p>
+Indexed and streaming search differ in execution
+and the match-configuration differs.
+Some times it is useful to inspect generated configuration
+to understand or validate the match configuration.
+Run this to find the value of the -i argument below:
```
jobergum, line 216: I don't like this section, it would be better to illustrate with an example instead of using low level shell commands to view generated configuration.

documentation/text-matching-ranking.html
```diff
@@ -207,6 +207,69 @@ <h3 id="prefix-search">Prefix Search</h3>
...
+ilscript[0].content[2] "clear_state | guard { input year | summary year | attribute year; }"
+ilscript[0].content[3] "clear_state | guard { input category_scores | summary category_scores | attribute category_scores; }"
+ilscript[0].content[4] "input tags | passthrough tags"
+</pre>
+</p>
+
+
+<h3 id="streaming-search">Streaming Search</h3>
+<p>
+See examples in <a href="streaming-search.html">streaming search</a>.
```
geirst, line 256: Perhaps we should have similar examples for indexed search?

documentation/text-matching-ranking.html
```diff
@@ -207,6 +207,69 @@ <h3 id="prefix-search">Prefix Search</h3>
 
 
 
+<h2 id="match-configuration">Match Configuration</h2>
```
geirst, line 210: As it stands, I think we should rename this section to something like Debugging Match Configuration. I think this kind of debugging is for an active developer, not the end user.

kkraune:
yes, agree, will move the low-level to less prominent place. just needed to put it _somewhere_ for next time...

will also harmonize indexed and streaming search examples - thanks for feedback!

documentation/text-matching-ranking.html
kkraune: Will add more examples of different matching for indexed in next PR - added todo

documentation/text-matching-ranking.html
kkraune: moved to end as debug section. will add more match examples for indexed search in next PR

kkraune:
Hid the low-level stuff, Todo for more examples as requested by @jobergum - that goes in next PR if OK with you - I would like to split this in more PRs due to size, please review

documentation/text-matching-ranking.html
geirst: I don't see the "move to end" change. Have you pushed it?

kkraune: sorry my bad, pushed now. I'll merge now and follow up in next PR - thanks!

PR #968, opened 2020-04-30
link to paper

kkraune:


kkraune:
that is a great idea, we will just change the link then - thanks for good doc of feature, btw!

In next PR and example with HNSW settings would be good / link from https://docs.vespa.ai/documentation/reference/query-language-reference.html#nearestneighbor to https://docs.vespa.ai/documentation/reference/schema-reference.html#index

PR #977, opened 2020-05-13
Update reference doc for distance metric on a tensor attribute.

geirst:
@arnej27959 please review
@jobergum FYI


documentation/reference/query-language-reference.html
```diff
@@ -737,7 +737,9 @@ <h2 id="where">where</h2>
     <a href="../tensor-user-guide.html">tensors</a> with one indexed dimension,
     where the size of that dimension is equal to the dimensionality of the vector space.
     The document positions are stored in a tensor attribute, and the query position is sent with the query request.
-    Euclidean distance is used as the distance function and the exact nearest neighbors are returned.
+    Euclidean distance is used as the default
+    <a href="schema-reference.html#attribute-distance-metric">distance metric</a>
+    and the exact nearest neighbors are returned.
```
arnej27959, line 742: while here, we should update this sentence to say that the query operator may return approximate nearest neighbors.

geirst:
@arnej27959 PTAL

PR #984, opened 2020-05-26
first cut docker images and rpms

kkraune:
@vekterli  _vespa-document-statistics_ is this still in use / remove?


vekterli:
There are no system tests using it and I think the underlying statistics visitor hasn't been tested for ages. So I'd say "not in use".

kkraune:
all right. so we should just remove it then - thanks! I left it out of docs

PR #1020, opened 2020-07-03
Extract monitoring quick start

bratseth:
@Oracien or @kkraune please review

We want to keep the quick start short.

### DIFF
--- a/documentation/monitoring-with-grafana-quick-start.html
+++ b/documentation/monitoring-with-grafana-quick-start.html
@@ -1,0 +1,94 @@
+---
+# Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+title: "Monitoring with Grafana quick start"
+---
+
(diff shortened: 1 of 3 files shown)


Oracien:
Personally I don't feel like it pollutes. But if this is principally is how the documentation should be structured then I approve.

kkraune:
we went a bit back and forth with this - we originally started with two docs. I think in 2020 is is cooler to get started quickly with some awesome graphs, it gives a much better first impression - it looks nice.

pasting some commands and observe JSON is OK, but not awesome. It takes just a minute or two more, and you will walk away with a much better experience. So I would like to keep it as is, with the distinction we made with part one and part two

bratseth:
There are many things people want to put here because it is really important, but I think the most important thing is to keep this to the smallest number of steps you actually need to get started. For example, if you are a researcher, or a developer in a non-opsdev org., or somebody who is investigating the feature set, you don't care about monitoring.

PR #1029, opened 2020-07-28
files have moved, update doc

kkraune:


kkraune:
agreed, but I could not find out , there seems to be some magic somewhere where the network is referred to as only backend, using the CWD implicitly. it would be better without, agree. can improve in next pr

Oracien:
Understandable. Will fix it later.

PR #1033, opened 2020-07-29
Kkraune/getting started i

kkraune:
more work needed, like simplifying the guides and add more content, but it is a start - easier for newbies. will add more coming week


Oracien:
👍

PR #1039, opened 2020-08-05
add more troubleshooting

kkraune:
moved from elsewhere


kkraune:
sure. can add more to troubleshooting later. but not in scope for this pr, there is no changes to content, just a move

PR #1067, opened 2020-08-28
Remove sample app text search testing

lesters:
@kkraune @thigm85 Please review. 

There were 2 tests - one from documentation and one from sample apps - that essentially tested a similar application: text-search. The documentation tested the process of developing it, the sample-app tested the finished application. We'll only keep the documentation one for now.


kkraune (approved):
Thanks - less confusion!

PR #1075, opened 2020-09-14
use table format, add automate section

kkraune:


documentation/vespa-plugins.html
```diff
@@ -20,55 +20,94 @@
...
+<tr id="automate">
+  <th>Automate</th>
+  <td>
+    <p>
+    Add testing to automate deployments:
+    <ul>
+      <li><em>System Tests</em> for functional tests -
+        a system test is an integration test that tests the application API endpoints</li>
+      <li><em>Staging Tests</em> to validate the instance before and after the change,
+          i.e. two test runs with potentially different tests</li>
```
jonmv, line 93: Perhaps something to further distinguish this from the above — staging tests validate that state in the cluster is not broken by an upgrade of the Vespa runtime, or application code or config?

documentation/vespa-plugins.html
```diff
@@ -20,55 +20,94 @@
...
+      <li><em>Production Tests</em> to validate a production instance before upgrading the next instance</li>
+    </ul>
+    System and Staging tests require a (downscaled) full instance,
+    whereas Production tests should be read-only on existing production instances.
```
jonmv, line 97: Well, we don't support read access to production clusters in tests in Vespa Cloud (yet?).

kkraune: right, will fix

documentation/vespa-plugins.html
kkraune: maybe. It is not _that_ interesting for the user that we also test during vespa upgrades, in the context of this document. can add some more details on code/config

kkraune:
@jonmv ptal

jonmv (approved):

PR #1077, opened 2020-09-24
Suggest different convencience for creating DocumentAccess, and menti…

jonmv:
…on injection

@bratseth please review.


bratseth (approved):

jonmv:
FYI @kkraune

PR #1096, opened 2020-11-05
Clarify that streaming search does not support indexing script

jonmv:
@baldersheim please review.


kkraune (approved):
changes look good, also consider a note in http://docs.vespa.ai/documentation/indexing.html for the differences in chain setup for the indexing modes

jonmv:
Will do.

PR #1103, opened 2020-11-22
first cut injectable components reference

kkraune:
I decided to make a component reference, and based it on the component types doc, that was reference doc anyway - no changes to content here

the only new part is from the spreadsheet, with some minor tweaks. this needs much more work, but want to isolate into smaller commits as I test and fix.

I also need to integrate this with much other doc, left some todos, there will be more

as the doc is not worse than before I suggest we merge for now, to make later PRs more manageable


documentation/reference/component-reference.html
```diff
@@ -0,0 +1,383 @@
...
+  <a href="#components">Component</a> is the basic building block in the container.
+  A component may be injected into other components,
+  including all the different request handlers, document processors and so on mentioned above,
+  and will get a destructor like method invoked when no other component is using it any more.
+</p>
+
+
+<table class="table">
+  <tr id="server">
+    <th>Server</th>
```
bjorncs, line 56: FYI the documentation for `ServerProvider` is no longer relevant as jdisc standalone/core is no more.

bjorncs (approved):

documentation/reference/component-reference.html
kkraune: ok - will remove the _server_ component then

PR #1114, opened 2020-12-08
Document ZooKeeper in container

mpolden:
@hmusum

### DIFF
--- a/_data/sidebar.yml
+++ b/_data/sidebar.yml
@@ -126,6 +126,8 @@ docs:
         url: /documentation/bundle-plugin.html
       - page: Versioning in the Container
         url: /documentation/jdisc/component-versioning.html
+      - page: Using ZooKeeper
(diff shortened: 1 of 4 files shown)


_data/sidebar.yml
```diff
@@ -126,6 +126,8 @@ docs:
         url: /documentation/bundle-plugin.html
       - page: Versioning in the Container
         url: /documentation/jdisc/component-versioning.html
+      - page: Using ZooKeeper
+        url: /documentation/using-zookeeper.html
```
mpolden, line 130: @kkraune: Should this be `.md` in this case?

hmusum (approved):
Great!

_data/sidebar.yml
kkraune: markdown files are auto-converted to html, so using .html in the link is correct

PR #1115, opened 2020-12-09
Fix link for component types

mpolden:
@kkraune


kkraune (approved):
thanks! will look into deadlink checker so we catch these auto

PR #1120, opened 2020-12-14
Reindexing doc

jonmv:
@kkraune please review.


jonmv:
Any comments, @vekterli? I guess it could mention that write races should not be an issue, too.

vekterli:
It would be good to mention that the system is designed to prevent write races between reindexing and concurrent feed(*), but that this comes at the cost of possibly elevated write latencies to the subset of the document space that is currently being reindexed.

_(*) to the extent possible in a system without strong consistency 🦄_

kkraune (approved):
Great stuff! Can't wait to try this ...

PR #1136, opened 2021-01-12
reindexing examples

kkraune:
not complete, but I want to add simple cut'n paste examples here, with ideas of what you can do with it. imho ok to merge as is and add more as we go, but OK with feedback for more info/changes.

### DIFF
--- a/documentation/reindexing.html
+++ b/documentation/reindexing.html
@@ -19,6 +19,8 @@ and inspected at the
 <a href="/documentation/cloudconfig/deploy-rest-api-v2.html#reindexing">reindexing endpoint</a>.
 </p>
 
+
+
 <h2 id="when-to-reindex">When to reindex</h2> 
 <p>
 When deployment results in a change in the indexing pipeline of an application, this is
@@ -35,7 +37,9 @@ the config generation that introduced the change. Reindexing then commences with
 deployment of the application. 
 </p>
 
-<h3 id="reindexing-progress">Reindexing progress</h3> 
+
+
+<h2 id="reindexing-progress">Reindexing progress</h2>
 <p>
 Reindexing is done by a component in each content cluster that
 <a href="/documentation/content/visiting.html">visits</a> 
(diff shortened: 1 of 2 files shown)


documentation/reindexing.html
```diff
@@ -61,3 +65,92 @@ <h3 id="reindexing-progress">Reindexing progress</h3>
...
+        To clear a field's data:
+<pre>
+# before
+field artist type string {
+    indexing:      summary | index
+}
+
+# after
+field artist type string {
+    indexing: "" | summary | index
```
jonmv, line 90: Can you do that to document fields, though?

jonmv: It's also potentially much more effective to feed that particular partial update to all documents in the corpus.

jonmv (approved):

documentation/reindexing.html
jonmv: This is not supported for document fields.

documentation/reindexing.html
```diff
@@ -61,3 +65,92 @@ <h3 id="reindexing-progress">Reindexing progress</h3>
...
+search mydocs {
+
+    field songs type array&lt;string&gt; {
+        indexing: input song | to_array | index | summary
+    }
+
+    document mydocs {
+        field song type string {
+            indexing: index | summary
+        }
```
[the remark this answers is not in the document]
jonmv: I guess simple to use, but more expensive, is OK, as long as the total cost still isn't too high. Seems a few days of reindexing for one of our larger hosted corpuses.

kkraune:
Thanks for useful feedback! I'll make some changes to this PR, then merge

documentation/reindexing.html
kkraune: correct, it will throw a deployment error

documentation/reindexing.html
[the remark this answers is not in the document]
kkraune: I think you are right, not sure if you can ever write to the new songs field. I will find a better example

PR #1139, opened 2021-01-14
link works but current link checker does not assume extension

kkraune:
we use @freva 's link checker here, it does not assume extension I believe. We can start using the htmlproofer we do for other repos, for now work around adding extension


jonmv (approved):
Thanks!

PR #1141, opened 2021-01-14
elastic cleanup part I

kkraune:
The elastic doc shall cover the elasticity features and benefits of them, not much more - move parts of it off, and less confusion what is a content node. This doc will be linked from the vespa.ai site most likely so much be in good shape

kill ugly illustration, reuse

will remove more in next PR


bratseth (approved):
👍

PR #1145, opened 2021-01-18
collapse into one document

kkraune:
Both attribute docs were dominated by memory stuff so is much better as one doc, it is not too long. This also killed some redundant text

Added input from Geir on how a multivalue update in memory is

fyi @jobergum


documentation/attributes.html
```diff
@@ -41,47 +26,471 @@
...
+}
+</pre>
+    </td>
+  </tr>
+  <tr id="fast-access">
+    <th style="white-space:nowrap;">fast-access</th>
+    <td>
+      Use to increase feed rates by having replicas on all nodes in memory.
+      <a href="reference/schema-reference.html#attribute">fast-access</a> locks all replicas of the attribute in memory.
+      Als see <a href="performance/sizing-feeding.html">sizing feeding</a>.
```
geirst, line 75: Als -> Also.

documentation/attributes.html
```diff
@@ -41,47 +26,471 @@
...
+        }
+    }
+}
+</pre>
+    </td>
+  </tr>
+  <tr id="fast-access">
+    <th style="white-space:nowrap;">fast-access</th>
+    <td>
+      Use to increase feed rates by having replicas on all nodes in memory.
```
geirst, line 73: Consider being more explicit on when _fast-access_ makes sense using.

documentation/attributes.html
```diff
@@ -41,47 +26,471 @@
...
+  </tbody>
+</table>
+
+
+
+<h2 id="multivalue-mapping-enum-store">Multivalue Mapping and Enum Store</h2>
 <p>
-<a href="reference/schema-reference.html#attribute">fast-access</a> locks all replicas of the attribute in memory.
-See <a href="performance/sizing-feeding.html">sizing feeding</a> for details.
+  The attribute Multivalue Mapping and Enum Store has upper limits, limiting the number of values in an attribute.
```
geirst, line 402: Consider a sentence explaining what these components are used for: multivalue mapping storing arrays of values for array and weighed set types, and enum store storing all unique strings for all string attributes.

documentation/attributes.html
```diff
@@ -41,47 +26,471 @@
...
+  document stored on a node.
+</p><p>
+  The document meta store is an implicit attribute, and is
+  <a href="proton.html#lid-space-compaction">compacted</a> and
+  <a href="proton.html#attribute-flush">flushed</a>.
+  Memory usage for applications with small documents can be dominated by this attribute,
+  particularly for <em>store-only</em> and <em>streaming search</em> applications.
+</p><p>
+  The document meta store scales linearly with number of documents -
+  using approximately 30 bytes per document on disk.
```
geirst, line 460: Consider removing "on disk", as this can be interpreted as the document meta store uses 30 bytes (on disk instead of memory) per document.

geirst (approved):
👍

documentation/attributes.html
kkraune: good point! there is some more details in the linked-to doc

documentation/attributes.html
kkraune: yup! left as a todo for now

PR #1146, opened 2021-01-18
reorganize elastic docs

kkraune:
now the elastic doc is more about concepts - the how-to is moved to procedural docs. I need another pass at the elastic doc, but easier to do now.

sorry hard to review this, not really needed as content is simplified and moved, not much new - next PR will be more meaningful to read


bratseth (approved):
Great!

PR #1148, opened 2021-01-19
address todo for enum store / multivalue data structures

kkraune:
reorder, reformat, not much new text except Geir's note in the todo now resolved


geirst (approved):
👍

PR #1192, opened 2021-02-18
deploy using REST API

kkraune:
- eliminates the need to mount the app package into the Docker container
- complicates a little if the user deploys twice, as the session id is required. It is _possible_ to parse this from the json response, but will make the steps uglier and less informative. instead added a little note in doc on this
- there are a few more guides that deploy from inside the Docker container, can update once this is final


bratseth (approved):
Great!

PR #1195, opened 2021-02-23
fix links

kkraune:
the link checker in this repo does not assume extension, add .html 

Lets look at how to promote this tutorial better, too


lesters (approved):
Thanks!

PR #1199, opened 2021-03-02
Component reference

bratseth:
Some corrections and improvements

### DIFF
--- a/en/reference/component-reference.html
+++ b/en/reference/component-reference.html
@@ -3,118 +3,62 @@
 title: "Container Component Reference"
 ---
 
-
-<h2 id="component-types">Component Types</h2>
 <p>
-  The container has a large number of different types of plug-ins available to a developer.
-  To start creating components, see the <a href="../developer-guide.html">Developer Guide</a>.
-</p><p>
-<img src="../img/container_components.svg">
-</p><p>
-  A client may or may not be a network client, so it is related to the network layer.
-  The network layer associates a Request with a <a href="#response-handlers">response handler</a>
-  and routes it to the correct type of <a href="#request-handlers">request handler</a>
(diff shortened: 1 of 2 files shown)


kkraune (approved):
Great cleanup!

en/reference/component-reference.html
```diff
@@ -3,182 +3,103 @@
...
-      </p><p>
-        A content channel is the basic I/O building block of JDisc.
-        It is conceptually similar to a stream,
-        but the basic unit is the ByteBuffer instead of the native byte.
-        A content channel will take ownership of a ByteBuffer written to it.
+        Request handlers lets applications implement arbitrary HTTP API's.
       </p>
-    </td>
-  </tr>
-  <tr id="response-handlers">
```
kkraune, line 81: Not sure if Response handler is a valid concept anymore / should be documented - ref https://github.com/vespa-engine/documentation/blob/master/en/reference/component-reference.html#L203 (broken link) and the illustration that still uses this name. Is it correct to just use the term Request handler everywhere?

bratseth: They are a valid concept for but not a component type, which made this confusing since it purports to be a list of component types. However, later in this I gave up trying to make this an actual list of components.

They are only relevant for people who want to write their own async handlers in jDisc, which is very far from what should be covered in a "component reference". The real problem here is that this documents is, in addition to containing info on components, is trying to teach about the working of jdisc core. 

I ran out of steam before cleaning that up completely, and I think. it's ok to leave it like it is, but if you want to bring the response handler doc back I guess that's fine too.

kkraune: Thanks! I will just remove the broken link for now - this doc is much better than it used to be :-)

PR #1203, opened 2021-03-05
link to code for memory estimate

kkraune:


kkraune:
@geirst

en/streaming-search.html
```diff
@@ -181,16 +181,10 @@ <h2 id="memory-sizing">Memory sizing</h2>
...
-<pre>
-&lt;tuning&gt;
-  &lt;searchnode&gt;
-    &lt;resizing&gt;
-      &lt;initialdocumentcount&gt;5000000&lt;/initialdocumentcount&gt;
-</pre>
-Note: This is a hard limit if the node does not have memory to keep more than one attribute in memory!
+</p>
+<p>
+See <a href="https://github.com/vespa-engine/vespa/blob/3b62cc5a2e7393c9399e7e8bda05085a91fc2ecc/config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java#L20">
```
geirst, line 186: I am not sure we should link to this at all. It is very low-level to make any sense for the user.

geirst (approved):
👍

en/streaming-search.html
kkraune: right. the question comes up now and then for such apps - how many docs can we estimate per node. Maybe we should just say "estimate 50 bytes per doc"?

geirst: Yes, I think that is better.

PR #1204, opened 2021-03-10
Make it explicit that this is for self-hosted

bratseth:
We've had a few cases where people have asked about this in relation to Vespa Cloud.


vekterli (approved):
:+1:

PR #1212, opened 2021-03-17
Add note on docker stop for graceful shutdown

jobergum:
It would be great if we can could make a pass and link to running docker in production from all sample apps which uses docker?


aressem (approved):

kkraune (approved):
Great!

kkraune:
yes we should link to this from all such sample apps, @jobergum  - I will put this on my todo list

PR #1215, opened 2021-03-19
draft getting started with ranking in Vespa

kkraune:
We try to make it easier for people try kick Vespa's tires, and we see that deploying something to try it out makes it harder / people drop off

we have set up https://doc-search.vespa.oath.cloud/ open for everyone, and I have added simple, sample rank-profiles there - with this guide, one can run queries with different profiles, to explore different things.

aimed at beginners, but I think we should add more to this so we can cut stuff from other docs and instead just refer to this live setup.

WIP, but want your opinion before putting more into it

@jobergum @lesters

### DIFF
--- a/en/getting-started-ranking.md
+++ b/en/getting-started-ranking.md
@@ -1,0 +1,267 @@
+---
+# Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+title: "Getting started with ranking"
+---
+
+Learn how [ranking](ranking.html) works in Vespa by using the open [query API](query-api.html) of
+[vespa-documentation-search](https://github.com/vespa-engine/sample-apps/tree/master/vespa-cloud/vespa-documentation-search).
+
+Ranking is the user-defined computation that scores documents to a query,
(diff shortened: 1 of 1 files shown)


jobergum:
This is a great start Christian.  

 I think that we need some more examples especially around using the rank query operator and also the wand and weakAnd operator.  also a document search so for example you can demonstrate using you search query and how you connect the terms that are used to has been typing if you use type all which is the default for the simple query language or you use type any and how that changes the retrieval.

 I will love to see something about an efficient retrieval function, for  example using approximate nearest neighbour search or using the wand and also our efficient range queries. Select the top-k  top-k popularity or the top-k freshest documents new start as a efficient retrieval functions and the hits retrieved in the retrieval function is then exposed to the first face ranking period

we get questions on how to do feel level weighting also would like to see traditional BN25 and boosting, ranking query operators. 

Note in the example you start changing keep bank account I'm not sure if that is a good idea I also note that the rank score drop limit is not an efficient way of dropping documents because you already retrieve them and you score them and then you drop them so it's a brute-force over the retrieved documents so I'm not sure if that's a great example especially in the context of multiphase ranking.

kkraune:
thanks for voice-to-text feedback! as agreed on video, lets merge and iterate next week on next items to add

PR #1224, opened 2021-04-09
update illustrations

kkraune:
the not-ready DB is now in both illustrations = less confusion

@bratseth or @jobergum


bratseth (approved):
👍

PR #1235, opened 2021-04-14
tensors, distance metric, partial update

kkraune:
@geirst


en/attributes.html
```diff
@@ -123,10 +141,19 @@ <h2 id="data-structures">Data structures</h2>
       One large memory buffer holding references (32 bit) to a memory structure
       with objects per unique value.
     </td>
+  </tr><tr>
+    <td>Tensor</td>
+    <td>Fixed / Variable</td>
+    <td>
+      Like the "D" attribute - a tensor is a multivalued datatype with sparse and/or dense dimensions,
```
geirst, line 148: We should use the terms "mapped and indexed dimensions" to be aligned with https://docs.vespa.ai/en/reference/tensor.html. I am also not sure we should say its a multivalue datatype.

en/attributes.html
```diff
@@ -123,10 +141,19 @@ <h2 id="data-structures">Data structures</h2>
...
+      stored in the <em>tensor store</em>.
+      Dense dimensions resemble the attribute multivalue mapping, as elements are fixed in size and numbers.
```
geirst, line 150: I would skip the comparison with the other data stores, as I find it a bit misleading. Consider the following simplification instead: "The memory layout in the tensor store depends on the tensor type".

geirst (approved):
👍

en/attributes.html
kkraune: "mapped and indexed dimensions" better, yes. removed multivalue

kkraune: not sure what we put in multivalue semantics, though, as a tensor has multiple values, but nevermind

en/attributes.html
kkraune: ok!

geirst (commented):
👍

PR #1240, opened 2021-04-16
Extract tensor examples and add more

bratseth:


en/tensor-examples.html
```diff
@@ -0,0 +1,246 @@
+---
+# Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+title: "Tensor Computation Examples"
+---
+
+<p>It is well know that tensors can be used to express machine-learned models such as neural nets,
+but they can be used for much more than that. The tensor model in Vespa is especially powerful,
+since it supports sparse dimensions, dimension names and lambda computations. Whatever you want
```
kkraune, line 8: per @geirst we should use "mapped and indexed" dimensions as names (i.e. not sparse). However, here, there property of the tensor is being sparse (a mapped tensor is both sparse and uses names), so OK.

en/tensor-examples.html
```diff
@@ -0,0 +1,246 @@
...
+
+<h2 id="dot-product-between-query-and-document-vectors">Dot Product between query and document vectors</h2>
+
+<p>Assume we have a set of documents where each document contains a vector of size 4.
+We want to calculate the dot product between the document vectors and a vector passed down with the query
+and rank the results according to the dot product score.</p>
+
+<p>The following schema file defines an attribute tensor field
+with a tensor type that has one indexed dimension <code>x</code> of size 4.
+In addition we define a rank profile that calculates the dot product.</p>
```
kkraune, line 92:
```suggestion
In addition we define a rank profile that calculates the dot product:</p>
```

en/tensor-examples.html
```diff
@@ -0,0 +1,246 @@
...
+<p>One simple way to use machine-learning is to generate cross features from a set of base features
+and then do a logistic regression on these. How can this be expressed as Vespa tensors?</p>
+
+<p>Assume we have three base features:</p>
+
+<pre>
+query(interests): tensor(interest{}) - A sparse, weighted set of the interests of a user.
+query(location): tensor(location{}) - A sparse set of the location(s) of the user.
+attribute(topics): tensor(topic{}) - A sparse, weighted set of the topics of a given document.
+</pre>
```
kkraune, line 162: these should be called mapped tensors

kkraune (approved):
This is great!

en/tensor-examples.html
geirst: Some reference: https://docs.vespa.ai/en/reference/tensor.html. Here we lay the foundation where a tensor dimension is either mapped or indexed. The tensor as a whole can be sparse (only mapped dimensions), dense (only indexed dimensions), mixed (a combination).

en/tensor-examples.html
bratseth: Here I call them sets. A tensor dimension is mapped or indexed. A tensor can be a map, set, matrix, vector etc, and can be sparse, dense, mixed etc.

PR #1241, opened 2021-04-19
remove symlink

kkraune:
@bratseth or @frodelu 

its been 3 months and google has picked up the /en/ docs (at leat the ones I just checked) and also ranks them higher


bratseth:
Isn't it better to just leave this in place?

kkraune:
if is has value, we can leave it, it is a bit confusing to me (but only in my editor), but not important

PR #1269, opened 2021-04-29
mention int8 and bfloat16 cell types

arnej27959:

@havardpe please review

### DIFF
--- a/en/tensor-user-guide.html
+++ b/en/tensor-user-guide.html
@@ -180,8 +180,10 @@ If memory (or memory bandwidth) is still a concern, it's possible to
 change the most space-consuming tensors to use the
 <code>bfloat16</code> cell type.  This type has the range as a normal
 32-bit float but only 8 bits of precision, and can be thought of as
-"float with lossy compression".  Some careful analysis of your data is
-required before using this type.
+"float with lossy compression".  See also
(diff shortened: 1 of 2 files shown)


havardpe (approved):
looks good

maybe drop saying that we will automatically cast to 32-bit float (results of calculations will be as if it was a 32-bit float)

(either) maybe refer to cell_cast operation as a way to ensure tensors have the same cell type to enable back-end optimization.
(or) probably better: we add the needed cell_casts automatically in the back-end.

en/tensor-user-guide.html
```diff
@@ -168,13 +168,48 @@ <h2 id="performance-considerations">Performance considerations</h2>
 Even though you would typically not do the above without reducing the
 search space first (using matching and first phase), it is important to consider
 the memory bandwidth and other hardware limitations when developing ranking expressions with tensors.
-</p><p>
+</p>
+<h3 id="cell-value-types">Performance considerations for cell value types</h2>
```
havardpe, line 172: header mismatch h2 vs h3

havardpe (approved):
looks good

arnej27959:
@kkraune please take a look

kkraune (approved):
Great!

PR #1285, opened 2021-05-07
Document sorting degrading

bratseth:


kkraune (approved):
Great! I will consider linking here from https://docs.vespa.ai/en/graceful-degradation.html as it is related

PR #1298, opened 2021-05-14
Part 10 of the news recommendations tutorial - models hot swap

erangi:
This PR adds a tutorial demonstrating techniques to hot swap recommendation models, namely apply updates with no downtime and no correctness violations. 


kkraune:
I will review this, I assume it is quite the same as in the shared doc. @lesters and I discussed whether this should be a part 10 or a more separate document, referring to the series (and the series could refer to this as a related subject) but not necessarily a part of it, as it is more about a different theme.

kkraune:
This looks great and I will merge it as-is into the Vespa Docs. Thanks for contributing!

I will however suggest we move it out of the series of news-recom and make this a standalone unit (although linking to this from the series). It has so many interesting subjects, that are also somewhat distinct, so we can maybe write a set of blog posts based on this, instead of only one. We can also add other aspects, making this into a series of its own.

To facilitate this, I will reorganize a little in the tutorials section first and add some links. Stay tuned for PRs, Eran!

erangi:
Thanks @kkraune!

PR #1323, opened 2021-06-04
expiry details

kkraune:
@vekterli or @jonmv please merge

### DIFF
--- a/en/documents.html
+++ b/en/documents.html
@@ -165,9 +165,9 @@ field timestamp type long {
 }
 </pre>
 <p>
-  When <code>garbage-collection="true"</code>, Vespa iterates of the document space to purge expired documents.
+  When <code>garbage-collection="true"</code>, Vespa iterates through the document space to purge expired documents.
   It is not possible to observe where in the cycle the process it,
-  and one can configure a pause between cycles using
+  but one can configure a pause between cycles using
   <a href="reference/services-content.html#documents.selection">garbage-collection-interval</a>.
   One can hence not know excatly when a document will be removed from the index by using expiry.
 </p>
--- a/en/documents.html
+++ b/en/documents.html
@@ -164,6 +164,13 @@ field timestamp type long {
     }
 }
 </pre>
+<p>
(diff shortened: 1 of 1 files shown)


en/documents.html
```diff
@@ -164,6 +164,13 @@ <h2 id="document-expiry">Document expiry</h2>
     }
 }
 </pre>
+<p>
+  When <code>garbage-collection="true"</code>, Vespa iterates of the document space to purge expired documents.
```
jonmv, line 168:
```suggestion
  When <code>garbage-collection="true"</code>, Vespa iterates through the document space to purge expired documents.
```

en/documents.html
```diff
@@ -164,6 +164,13 @@ <h2 id="document-expiry">Document expiry</h2>
...
+  It is not possible to observe where in the cycle the process it,
+  and one can configure a pause between cycles using
```
jonmv, line 170:
```suggestion
  but one can configure a pause between cycles using
```

jonmv (approved):

en/documents.html
```diff
@@ -164,6 +164,13 @@ <h2 id="document-expiry">Document expiry</h2>
     }
 }
 </pre>
+<p>
+  When <code>garbage-collection="true"</code>, Vespa iterates through the document space to purge expired documents.
+  It is not possible to observe where in the cycle the process it,
+  but one can configure a pause between cycles using
+  <a href="reference/services-content.html#documents.selection">garbage-collection-interval</a>.
```
vekterli, line 171: \*Pushes glasses up to nose\*

_Technically_ 🤓 the GC is always running in the background when enabled. Vespa distributes future GC deadlines for all buckets pseudo-randomly within the configured time interval (it chunks the wall-clock timeline into discrete, non-overlapping intervals of this size). The interval is therefore not a pause as per se, but the minimum time between each time a particular bucket (and therefore a particular document, assuming it stays in the same bucket) is considered for GC.

But it's definitely the case that you cannot know for sure when a document will be GC'd.

PR #1326, opened 2021-06-04
use zip instead of tar|gzip

kkraune:
ref https://github.com/vespa-engine/vespa/issues/17837 - some times issues with tar, work around for now

FYI @chunnoo @olemagnusnorum

### DIFF
--- a/en/reference/application-packages-reference.html
+++ b/en/reference/application-packages-reference.html
@@ -85,7 +85,8 @@ Use whatever name for the zip file - then refer to the file instead of the path 
 <a href="vespa-cmdline-tools.html#vespa-deploy">example</a>.
 <strong>Note:</strong> Using <code>tar</code> / <code>gzip</code> is not supported.
 </p>
-
+<!-- ToDo: reconcile this with cloudconfig/deploy-rest-api-v2.html
+     that says "Set Content-Type HTTP header to application/x-gzip or application/zip" -->
 
 
 <h2 id="deploy">Deploy</h2>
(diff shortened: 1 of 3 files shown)


en/reference/application-packages-reference.html
```diff
@@ -85,7 +85,8 @@
 <a href="vespa-cmdline-tools.html#vespa-deploy">example</a>.
 <strong>Note:</strong> Using <code>tar</code> / <code>gzip</code> is not supported.
 </p>
-
+<!-- ToDo: reconcile this with cloudconfig/deploy-rest-api-v2.html
+     that says "Set Content-Type HTTP header to application/x-gzip or application/zip" -->
```
kkraune, line 89: @hmusum / @jonmv I am not sure whether we support it or not

hmusum: Zip should work, tested here: https://github.com/vespa-engine/system-test/blob/master/tests/cloudconfig/deploy/deploy.rb#L268

hmusum (approved):

en/reference/application-packages-reference.html
kkraune: yes zip works - I am just not sure if tar | gzip is supposed to work, ref the deploy-rest-api-v2.html that says you an submit application:x-gzip

hmusum: tar and gzip with `Content-Type:application:x-gzip` is the default used by `vespa-deploy`, see https://github.com/vespa-engine/vespa/blob/master/config-model/src/main/perl/vespa-deploy#L447-L453 (zip is used if ´$app` is a file, otherwise tar and gzip is used)

kkraune: great! lets follow up in vespa-engine/vespa#17837 for the problematic tar archives

PR #1338, opened 2021-06-13
add how to run sub-queries to FAQ

kkraune:
link to @lesters example + skeleton for a scheduled refresher that we should maybe add to a sample app instead, add here for now


bratseth (approved):

bratseth:
👍

PR #1341, opened 2021-06-15
Remove ignored config option min-group-coverage

bratseth:


kkraune (approved):
can we just remove or should we leave note that it will be rmeoved in vespa 8? not sure if someone has used this. OK to remove if it has no effect, but we should leave a note if deploy will fail if someone open source uses it

bratseth:
It already has no effect and is ignored (no failure).

baldersheim (approved):

PR #1343, opened 2021-06-19
Document single value tensorFromLabels

bratseth:


bratseth:
Should we just remove the stuff about specifying query(property)? It's easier to just send a tensor.

arnej27959 (approved):
looks good to me.  sending a query property and using tensorFromLabels avoids having to declare the type in a query profile type, but I agree it might be easier to have just one way to send query tensors.

PR #1355, opened 2021-07-01
Kkraune/auto test ranking

kkraune:
Instead of `select * from sources * ` use `select * from doc` to restrict existing queries to the `doc` schema - so we can easily add more schemas later. Add auto testing to ensure this does not break


olemagnusnorum:
skal denne merges selv om den failer paa screwdriver?

kkraune:
yes please approve/merge, link check failures are unrelated

olemagnusnorum (approved):

PR #1356, opened 2021-07-01
Add doc for unit testing model evaluation

lesters:
@kkraune Please review.


kkraune (approved):
Thanks! It _could_ be that a few &lt;p&gt; 's here and there will make the rendered text look more like the source ;-)

lesters:
Thanks, too used to markdown at this point!

PR #1363, opened 2021-07-07
Kkraune/improve admin procedures

kkraune:
It is primarily the first commit that adds much value, the rest is cleanup / move text for next rewrite of this doc - the purpose is to make this more procedural, so we can link to from support requests


hmusum (approved):
Great!

PR #1373, opened 2021-08-04
Enable search suggest dropdown for all docs

kkraune:


olemagnusnorum (approved):

chunnoo:
I would have placed the <script> tag at the end of the file. It seems weird to have it inside a <div>.

PR #1375, opened 2021-08-06
Add procedure to change attribute aspect

kkraune:
FYI @geirst 

document the relatively common procedure to migrate from attribute to index (or the other way) safely, using the new reindex features

### DIFF
--- a/en/operations/admin-procedures.html
+++ b/en/operations/admin-procedures.html
@@ -121,6 +121,217 @@ $ <a href="../reference/vespa-cmdline-tools.html#vespa-model-inspect">vespa-mode
 
 
 
+<h2 id="schema-changes">Schema changes</h2>
+<p>
+  Refer to <a href="../reference/schema-reference.html#modifying-schemas">modifying schemas</a>
+  for reference documentation for making changes to a schema, and actions required.
+  Find procedures below for a subset of such changes.
+</p>
+
+
+<h3 id="attribute-to-index">Attribute to index</h3>
+<p>
+  Changing between <code>index</code> and <code>attribute</code> is a common field change operation
+  to optimize performance.
+  Use the <a href="../reindexing.html">reindexing</a> feature to safely migrate data to/from index structures.
+</p>
+<p>
+  When the attribute aspect of a field is removed, the attribute data is dropped on next process restart.
+  Hence, doing a direct change from attribute to index is risky, as a node stop can incur field data loss.
+  A safer approach is to run with both in the transition.
+  The gist of this procedure is hence to add <code>index</code>, run a reindex - then remove attribute aspect:
+</p>
+<pre>
+# field configuration at start
+field artist type string {
+    indexing: summary | attribute
+    attribute: fast-search
+}
+
+-&gt;
+
+# intermediate step to populate index structure, keeping the data in the attribute
+field artist type string {
+    indexing: summary | attribute | index
+    match: word
+    stemming: none
+}
+
+-&gt;
+
+# final configuration, migrated to index
+field artist type string {
+    indexing: summary | index
+    match: word
+    stemming: none
+}
+</pre>
+<p>
+  Procedure - test this using the <a href="../vespa-quick-start.html">quick-start</a>,
+  changing the <code>artist</code> field to an attribute before running.
+  Also add a <a href="../reference/validation-overrides.html">validation override</a>
+  in <code>src/main/application/validation-overrides.xml</code>:
+</p>
+<pre>
+&lt;validation-overrides&gt;
+    &lt;allow until="2021-08-30">indexing-change&lt;/allow&gt;
+&lt;/validation-overrides&gt;
+</pre>
+<p>
+  Run the quick start, keep the Docker container up.
+  Run a query to validate data can be queried -
+  also dump current index structures, see <code>artist</code> as an attribute:
+</p>
+<pre>
+$ curl "http://localhost:8080/search/?ranking=rank_albums&yql=select%20%2A%20from%20sources%20%2A%20where%20artist%20contains%20%22Coldplay%22%3B"
+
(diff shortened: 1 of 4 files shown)


en/operations/admin-procedures.html
```diff
@@ -121,6 +121,217 @@ <h2 id="status-pages">Status pages</h2>
...
+
+<h3 id="attribute-to-index">Attribute to index</h3>
+<p>
+  Changing between <code>index</code> and <code>attribute</code> is a common field change operation
+  to optimize performance.
+  Use the <a href="../reindexing.html">reindexing</a> feature to safely migrate data to/from index structures.
+</p>
+<p>
+  When the attribute aspect of a field is removed, the attribute data is dropped on next process restart.
+  Hence, doing a direct change from attribute to index is risky, as a node stop can incur field data loss.
```
jonmv, line 140: 😱

en/operations/admin-procedures.html
```diff
@@ -121,6 +121,217 @@ <h2 id="status-pages">Status pages</h2>
...
+</p>
+<pre>
+$ curl http://localhost:19071/application/v2/tenant/default/application/default/environment/prod/region/default/instance/default/reindex
+
+{
+  "generation": 3,
+  "applicationPackageFileReference": "",
+  "modelVersions": ["7.423.37"]
+}
+
```
jonmv, line 257:
```suggestion
$ curl http://localhost:19071/application/v2/tenant/default/application/default/environment/prod/region/default/instance/default/reindex

{
  "generation": 3,
  "applicationPackageFileReference": "",
  "modelVersions": ["7.423.37"]
}

```

This seems to match something completely irrelevant? That path is used with POST/DELETE to (de)activate reindexing, AFAIR, but isn't documented.

en/operations/admin-procedures.html
bratseth: Not really data loss though, but a gap in availability?

en/operations/admin-procedures.html
```diff
@@ -121,6 +121,217 @@ <h2 id="status-pages">Status pages</h2>
...
+      },
+      "ready": {
+        "music": {}
+      }
+    }
+  }
+}
+
+$ (cd src/main/application && zip -r - .) |   curl --header Content-Type:application/zip --data-binary @- \
+  localhost:19071/application/v2/tenant/default/prepareandactivate
```
jonmv, line 274: Hmm, why not use `vespa-deploy`? I thought that was the most common use? The `/application/v2/.../reindexing` API was intended for manually triggering reindexing, when using that deploy tool. `prepareandactivate` indeed lets the change be automatically detected, and acted upon, but I don't think that's even in the opensource documentation—neither of them?

jonmv: ... not saying it shouldn't be, though.

en/operations/admin-procedures.html
```diff
@@ -121,6 +121,217 @@ <h2 id="status-pages">Status pages</h2>
...
+      }
+    }
+  }
+}
+
+$ docker exec vespa /usr/bin/sh -c '/opt/vespa/bin/vespa-logfmt | grep Reindexer'
+[2021-08-06 09:13:00.524] INFO    : container-clustercontroller Container.ai.vespa.reindexing.Reindexer
+                                    Completed reindexing of datatype music (code: 1412693671) after PT0.547121S
+
+$ docker exec vespa /usr/bin/sh -c '/opt/vespa/bin/vespa-proton-cmd --local triggerFlush' && \
```
jonmv, line 298: What/why is this?

jonmv (commented):
Much great information here. I do have some comments, though.

en/operations/admin-procedures.html
kkraune: will test again to qualify this, could be I did this wrong

kkraune:
talked with Jon Marius, clarified a lot, v2 of this procedure coming shortly

en/operations/admin-procedures.html
kkraune: Using `prepareandactivate` makes it more similar to Vespa Cloud, easier to swap between the two. See https://docs.vespa.ai/en/cloudconfig/deploy-rest-api-v2.html#prepareandactivate . This doc needs to be updated as it is more than a prepare/activate (as it actually will do reindex), will do this soon

en/operations/admin-procedures.html
kkraune: I added this to be able to see the effects of the indexing changes - I think I will move this to an appendix instead, for debugging purposes, not really needed here

kkraune:
@jonmv updated after we talked yesterday, the procedure now works without the reindex call that was broken anyway.

I will separate this procedure into an own doc and add some more details / have a separate troubleshooting section etc. But would like to merge this in for now.

Finally, will test again to see if the node actually drops data or just temporarily

jonmv (approved):
Nice!

en/operations/admin-procedures.html
kkraune: ok - so maybe we need a new deploy api command then as https://docs.vespa.ai/en/cloudconfig/deploy-rest-api-v2.html requires a session (unless one can use * or something as session ID to use latest / create one auto)..

We can discuss that separately, I'll create a ticket

jonmv: > activate without prepare does an implicit prepare

I don't think that's true, for the HTTP API—looking at the code, the activate endpoint explicitly creates a `Deployment.prepared(...)`, which is then activated.

en/operations/admin-procedures.html
kkraune: you are right, @bratseth  - the data survived a proton restart even though not yet reindexed from attribute into index. after subsequent deployment, which triggered the actual reindex, the index was populated. Will rewrite to availability not loss in next PR. Thanks!

PR #1396, opened 2021-08-24
Remove <style> elements from search

kkraune:
This will fix search results like
```
... right: 30px;} The Vespa access log
```

Thanks for reporting, @jobergum !


jobergum (approved):

jobergum:
Nice!

PR #1407, opened 2021-08-29
Specify that you may inherit rank profiles from parents

bratseth:


kkraune (approved):
Thanks! I'll expand a little in https://docs.vespa.ai/en/getting-started-ranking.html on this as well, so users can find it earlier, it is very useful to know

PR #1418, opened 2021-09-01
Kkraune/schema

kkraune:
The schema guide is now more high level and with less irrelevant text

### DIFF
--- a/en/schemas.html
+++ b/en/schemas.html
@@ -196,6 +196,37 @@ $ENDPOINT/search/?
 
 
 
+<h2 id="schema-modifications">Schema modifications</h2>
+<p>
+  Vespa is built for safe schema modifications,
+  like adding a field or changing indexing or match modes.
+  A new version of the schema is deployed in an <a href="cloudconfig/application-packages.html">application package</a>.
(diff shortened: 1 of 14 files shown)


kkraune:
@hmusum please review

en/schemas.html
```diff
@@ -257,5 +255,32 @@ <h2 id="content-cluster-mapping">Content cluster mapping</h2>
...
+</pre>
+<p>
+  To accept such changes, add a <a href="reference/validation-overrides.html">validation-override</a>, e.g.
+</p>
+<pre>
+&lt;validation-overrides&gt;
+    &lt;allow until="2021-08-30"&gt;indexing-change&lt;/allow&gt;
+&lt;/validation-overrides&gt;
+</pre>
+<p>
```
hmusum: I don't think a change leading to a restart is a good example, the change will happen anyway and there is no validation override for restart

hmusum (changes requested):

en/schemas.html
kkraune: Good point! modifying to a field indexing change instead

kkraune:
@hmusum  ptal

hmusum (approved):
Great!

PR #1419, opened 2021-09-01
Link to search suggest sample app

kkraune:
I will take a stab at https://github.com/vespa-engine/sample-apps/blob/master/incremental-search/search-suggestions/README.md to cover more details and design issues - like how to use an array of strings to recommend _ranking_ from the _learning_to_rank_ suggestion ... etc.


kkraune:
@hmusum please review

hmusum (approved):

PR #1420, opened 2021-09-02
Revert "Revert "Tensor playground: post as form, not json body to avoid preflight""

lesters:
Reverts vespa-engine/documentation#1417

@kkraune Please merge.


kkraune (commented):
👍

PR #1424, opened 2021-09-07
Kkraune/partial updates

kkraune:
We can possible make it _even clearer_ why partial updates is a great feature ...


jobergum (approved):

jobergum:
Great stuff!

PR #1448, opened 2021-09-23
minor fixes

kkraune:
- https://docs.vespa.ai/en/reference/vespa-cmdline-tools.html should mention the vespa cli, and linked from here

### DIFF
--- a/en/vespa-quick-start-cli.html
+++ b/en/vespa-quick-start-cli.html
@@ -39,7 +39,7 @@ Make sure you see at minimum 6Gb.
 <li>
 <p><strong>Install the Vespa CLI:</strong></p>
 <pre data-test="exec">
-brew install vespa-cli or similar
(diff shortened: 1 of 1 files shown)


en/vespa-quick-start-cli.html
```diff
@@ -39,7 +39,7 @@ <h2 id="running-vespa-in-docker">Running Vespa in Docker</h2>
 <li>
 <p><strong>Install the Vespa CLI:</strong></p>
 <pre data-test="exec">
```
mpolden, line 41: This probably shouldn't be executed by the documentation tests, as that environment doesn't have `brew`.

kkraune: good point, will remove - the binary is installed in the image used afaik

mpolden (approved):

PR #1453, opened 2021-09-27
Wait with curl until next CLI release

mpolden:
@aressem


bratseth (commented):
Can we just skip the --wait arg?

mpolden:
Not without doing something else to make doc tests wait for configserver to become ready.

bratseth:
I suggest using the "hidden" trick.

bratseth (approved):

PR #1455, opened 2021-09-29
Kkraune/bundle reload

kkraune:
only the second commit needs review - it is a little slim, but useful as it is, feel free to add to it as needed


gjoranv (approved):
Great, thanks!

PR #1467, opened 2021-10-06
Add documentation for tensor expand non-primitive function

lesters:
@kkraune Please review.

Ref: https://github.com/vespa-engine/vespa/pull/19430


kkraune (approved):
Great!!

PR #1472, opened 2021-10-07
Improve doc

bratseth:
I'm no fan of this kind of "inductive" documentation, but it *did* try to say what all() did, at the top.
Anyway, I've tried to improve it a bit without going too far, please take a look.

### DIFF
--- a/en/grouping.html
+++ b/en/grouping.html
@@ -30,7 +30,7 @@ list of all matches to the query somewhere would be too expensive, and execution
   <li><code>output</code>: Output some value(s) at the current location in the structure.
 </ul>
 
-<p>The parallel and nested collection of these operations define both the structure of the computation
+<p>The parallel and nested collection of these operations defines both the structure of the computation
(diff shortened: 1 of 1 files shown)


en/grouping.html
```diff
@@ -8,26 +8,48 @@
...
+<p>The operations defining the structure of a grouping are:
+
+<ul>
+  <li><code>all(statement)</code>: Perform the nested statement once on the input list as a whole.
+  <li><code>each(statement)</code>: Perform the nested statement on each element of the input list.
+  <li><code>group(specification)</code>: Turn the input list into a list of list according to the grouping specification.
+  <li><code>output</code>: Output some value(s) at the current location in the structure.
+</ul>
+
+<p>The parallel and nested collection of these operations define both the structure of the computation
```
kkraune, line 33:
```suggestion
<p>The parallel and nested collection of these operations defines both the structure of the computation
```

PR #1476, opened 2021-10-08
Document query parameter ranking.matching.

baldersheim:
@kkraune or @bratseth or @jobergum PR


kkraune (approved):
This and the previous PR is great reference doc! next step is looking at the perf doc and add some links I suppose

PR #1543, opened 2021-11-05
Get less hits

kkraune:
- lets see how much faster then

### DIFF
--- a/js/handle_results.js
+++ b/js/handle_results.js
@@ -1,12 +1,9 @@
 const handleResults = (data) => {
   const result = document.getElementById("result");
-
   result.innerHTML = "";
 
   const hits = data;
   if (hits && hits.length > 0) {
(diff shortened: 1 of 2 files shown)


frodelu (approved):
OK with me, but wouldn't this mess up pagination? So we'll show only the first 25 results?

kkraune:
good point. we don't have pagination, btw. 25 hits vs 250 hits is approx 500ms faster. need to make more changes, too, as #hits is not right displayed. let me think more

kkraune:
more complicated than I thought, as the search recommendations are mixed into this as well. so "hits" displayed is just a count of all hits in documentation, which is only correct if we always ask for more hits than docs ...

number of hits is not that important, so lets just drop if for now and experiment with getting less hits - and think of adding pagination if we need to

kkraune:
so lets test this - should be snappier. look at pagination if we need to later

PR #1579, opened 2021-11-23
Add clipboard copy button to pre elements

kkraune:
@bratseth or @frodelu 

copy

works in Chrome, Firefox and Safari for me.

I have not added a visual indicator when it is clicked, but lets see if this is a step in the right direction


bratseth (approved):
Great

PR #1587, opened 2021-11-24
Add Query image

kkraune:
This adds the smaller image at the end of proton.html. Note that with the default size, the image renders too large, imho. There are many ways to set dimensions, ref https://css-tricks.com/scale-svg/ - but we should keep it simple, as many people will write docs / create illutrations.

Lucid does not add the viewbox attribute - we _could_ do it, but it is a hassle to open the svg after export and modify it, and must redo for each export

setting the width in pixels on the &lt;img&gt; to half of the pixels in the svg makes it about right - merge this and evaluate. we should do this consistently over all images, so lets discuss approach and size before we do more illustrations

@oyving / @frodelu


oyving (approved):

oyving:
Don't know why the linkchecker is complaining. This change looks unrelated.

As for the width - we should probably use max-width or something similar so that we don't force a certain width on the figure for smaller screens.

kkraune:
good point. I did not see any difference using width / max-width on my laptop - lets merge and see how the Iphone renders this. it is a little more cumbersome to use a style attribute vs width, so if there is no real change we can keep width

PR #1593, opened 2021-11-30
Remove documentation for 'servlet' in container.

gjoranv:
- The servlet integration has been not been used for years.

FYI: @bjorncs, @bratseth


kkraune (approved):
Thanks!

PR #1650, opened 2021-12-08
Combine two docs of same subject for readability MERGEOK

kkraune:
- some style fixes + typos, too
- no real changes to content


gjoranv:
The whole chain-builder.html doc can be removed. The ChainBuilder that is described is not even used internally. I will add it to the list of deprecated classes to be removed in Vespa 8.

kkraune:
Great! will remove it in another PR then

PR #1693, opened 2021-12-17
Run multiple guides in one test run

kkraune:
Started with the first two in the news series, will add the others in subsequent PRs

This makes is much easier to test from the actual steps in the doc instead of the hidden test elements + saves much time / resources


lesters (approved):
Great!

PR #1694, opened 2021-12-17
Test part 3 and 4 in news tutorial

kkraune:
- increased timeout, the generation of files takes time. will see how long on the auto test system
- typos, format


lesters (approved):
:+1:

PR #1723, opened 2022-01-03
Remove service element

gjoranv:
Remove doc for the generic 'service' element.

FYI: @hmusum


kkraune:
Thanks! I hope/think no-one is using this, so should be safe to remove doc / we can always put it back if someone complains. Normally we would keep doc with deprecation warning until next major. Merge if OK, @bratseth

kkraune (approved):

PR #1729, opened 2022-01-05
Set MathJax config, then load script. Smaller equations

kkraune:
- Equations look better with somewhat smaller text, now easy to configure.
- Using SVG instead of CommonHTML - will see how it looks on smaller screens / load times.

The table in reference/rank-features.html is still too wide, need to debug more to find the offending element


bratseth (approved):
👍

PR #1737, opened 2022-01-06
Add more files to sidebar / TOC

kkraune:
- minor fixes to json, text, format, too


bjormel (approved):
Is it the redirects that is causing the link checker to fail?

kkraune:
the link checker fails (correctly) on moved pages. I add a redirect from old location (the last one below) when moving files like

    redirect_from:
    - /documentation/node-setup.html
    - /en/node-setup.html

so end users can still use the old link / not lost in web search. I will correct the links before merging, thanks for checking!

bjormel:
Saw that a tad too late, will undo

PR #1742, opened 2022-01-07
Document plain not

bratseth:


kkraune (approved):

arnej27959 (approved):
As reference this looks good, I hope we still guide people to ANDNOT as the recommended operator.

bratseth:
Yes

PR #1744, opened 2022-01-09
Document weakAnd option

bratseth:


en/reference/simple-query-language-reference.html
```diff
@@ -16,9 +16,17 @@
 <table class="table">
   <tr>
     <th>all</th>
-    <td>Default. All the words of the query must match a
-        document for it to be a match.</td>
-  </tr><tr>
+    <td>Default. Created an AND: All the words of the query must match.</td>
```
arnej27959, line 19: Created -> Creates

arnej27959 (approved):
👍

en/reference/simple-query-language-reference.html
bratseth: Fixed, thanks

PR #1761, opened 2022-01-14
Simplify the description of now

bratseth:
Afaik the restrictions mentioned here were lifted long ago.


vekterli (approved):
:+1: I seem to recall GC used to be run as a search in ye olde days, so this restriction probably dates back to that

bratseth:
Yep

PR #1782, opened 2022-01-24
Kkraune/minikube

kkraune:
- simplified by not mounting a volume, deploy like the other guides

I will automate testing in a later PR, this makes the guide work at least


kkraune:
... or @hmusum

hmusum (approved):

PR #1792, opened 2022-01-26
Kkraune/metrics

kkraune:
- Add illustrations for better understanding of the different APIs and endpoints
- rename headings to same as API for clarity

The content of the doc is more or less unchanged, I wanted a review of the structure and naming before making other changes. It is OK to merge and read it, and make a new PR for next changes, or revert this


kkraune:
btw: if this is OK, I will rename the anchor texts from the other documents, too

gjoranv (approved):
This looks good. But beware that the term "Node metrics API" is used e.g. in the DataDog docs and error messages:

https://docs.datadoghq.com/integrations/vespa/#service-checks

https://github.com/DataDog/integrations-extras/blob/master/vespa/datadog_checks/vespa/vespa.py#L44

kkraune:
> But beware that the term "Node metrics API" is used e.g. in the DataDog docs and error messages:

good point! I will have that in mind for next PR

PR #1798, opened 2022-01-28
Kkraune/metrics cleanup

kkraune:
- with this, the structure of the document should be correct (moved under /state/v1/metrics)
- now ready to add doc of parameters, where applies


gjoranv (commented):
Great, except for commit: "Custom metrics are on /state/v1/metrics". The main entry point for metrics is "metrics/v2", and this is also where users typically will find their metrics after adding a custom consumer via services.xml. This is currently lacking in the metrics doc, but mentioned in the services.xml/admin reference.

kkraune:
Good point! I will rearrange the doc to focus on metrics/v2, you are right, this is where users will normally get metrics - the rest is just for reference

PR #1806, opened 2022-02-02
Remove internal dev docs about config from the sidebar.

gjoranv:
FYI: @hmusum


kkraune:
I agree that we should not have internal docs in the sidebar. We have also had requests that _all_ docs should be navigable. so maybe these docs belong elsewhere?

gjoranv:
internal docs in a ´legacy/7/´ folder perhaps?

kkraune:
we do that for the company internal docs, so an option. but this is not vespa 7 legacy, it is current? I am just pondering whether we should have a folder for vespa developers (open source) where we put such docs. I'll raise the discussion in the ticket meeting today

gjoranv:
If these docs are not maintained, they will become legacy. For example, many types will move to a new Java package for Vespa 8. I'm not sure it's worth the effort to maintain them. @hmusum ?

hmusum:
I think parts of this is worth keeping as developer doc, but some can be internal only (e.g. the C++ doc will be used by vespa developers only), let's discuss in ticket meeting.

kkraune:
@gjoranv  or @hmusum  please merge

hmusum (approved):

PR #1811, opened 2022-02-02
more geo position changes

arnej27959:

Documents change in https://github.com/vespa-engine/vespa/pull/20994
we also want to drop the "PosSearcher" that picks up "pos.ll" etc from the standard vespa chain.

@bratseth please review
@kkraune @gjoranv FYI


bratseth (commented):
👍

PR #1820, opened 2022-02-04
Kkraune/annotations reference

kkraune:
To review, do by each commit, the first is just rearranging. This gist of it is, create one table (with more columns) of annotations, add the missing annotations with ToDos, so next PR can move the reference doc into this table (many annotations are documented elsewhere or missing) and link here instead. Also, link to to guides from here for examples.

Maybe easier to review once I am fully done, this makes it a little better with few changes to actual content. This is a mess now ...


bratseth (approved):
Great

PR #1822, opened 2022-02-07
Document more YQL annotations

kkraune:
- the gist of it is, document the annotation's behaviour once in a table of annotations, link to it from one or more places used
- still a few annotations to go
- few changes to content, other than re-arranging and adding the obvious. Will address some todos later once I know more
- need another pass and also look at the other docs linking into this and refine the annotation description / link out. the yql guide can be made much better once this is in place ...

Question: is it important to separate annotations valid for string literals and/or functions? There was a distinction in the old docs, but I don't see that is is useful for the user reading this, would like to remove a column


kkraune:
I also assumed near and onear have the same annotation(s)

bratseth:
Great! Yes I don't think that distinction matters very much.

PR #1827, opened 2022-02-09
Document parmeter substitution

kkraune:
@jobergum 

I am not sure if _parameter substitution_ is the best name, open for suggestions. This PR defines the format once, and link from the three operators using it. Also, userInput uses the same mechanism, with free-text input


kkraune:
ping

bratseth (commented):
👍

PR #1831, opened 2022-02-11
Format default result format to match output

bartek:
This moves the list of fields typically found under the children array
of objects, more closely matching the expected JSON result format.

Additionally, adds a missing verb ("is") to one sentence.

Screenshot of formatting change

Screen Shot 2022-02-11 at 4 05 17 PM
:

### DIFF
--- a/en/reference/default-result-format.html
+++ b/en/reference/default-result-format.html
@@ -28,6 +28,51 @@ in other words, only strings are used as map keys.
             <li><code>children</code> optional, array of objects<br>
                 Array of JSON objects with exactly the same structure as
                 <code>root</code>.
+                <ul>
+                    <li><code>fields</code> optional, map of string to object<br>
+                        <ul>
+                            <li>The named document fields, this is where fields of search documents go</li>
+                            <li><a href="schema-reference.html#summary-features">summaryfeatures</a></li>
+                            <li><a href="schema-reference.html#match-features">matchfeatures</a></li>
+                        </ul>
+                    </li>
+
+                    <li><code>id</code> optional, string<br>
+                        String or URI identifying the hit, document or other data type.
+                    </li>
+
+                    <li><code>label</code> optional, string<br>
+                        The label of a grouping list.
+                    </li>
+
+                    <li><code>limits</code> optional, object<br>
+                        Used in grouping, the limits of a bucket in histogram style data.
+                        <ul>
+                            <li><code>from</code> optional, string<br>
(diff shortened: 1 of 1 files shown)


kkraune:
Thanks! I will double check and review on Monday

bartek:
@kkraune Thanks for looking! I am a bit uncertain if this is accurate, but it's what more closely resembles what I see in results from the engine. My only concern is the description of that node, _Array of JSON objects with exactly the same structure as `root`_ (It seems like the structure is similar, but not quite the same .. this might be a consequence of the particular setup I am using though)

kkraune:
Hi again! Yes this is very confusing, but I think technically not wrong as it is, as we can see from the json schema: 

````
        "document_node": {
            "properties": {
                "children": {
                    "type": "array",
                    "items": {
                        "type": "document_node"
````

But the recursive structure creates the confusion. I am hence inclined to keep the json schema, and just remove the list hierarchy (instead of your change), pointing to a query result example added to https://docs.vespa.ai/en/query-api.html instead. Most users will get by easily using the (added) example result, and those that need it, can use the json schema is it is.

This OK with you, @bratseth, remove the confusing list and add an example in the guide instead ?

bartek:
@kkraune Ah, makes sense now that I look at it through the lens of the JSON schema. My confusion primarily came from optional fields like `source` and `types` which I observed as being present in the children nodes, but not in the root node. Makes sense since they are optional. It's too bad JSON does not allow comments, otherwise the JSON schema would be enough to demonstrate the structure :)

Appreciate you clearing that up.

bratseth:
>  @bratseth, remove the confusing list and add an example in the guide instead ?

Yes we should really have an example, but are you suggesting to not have a list of the possible sub-elements and what they mean?

kkraune:
I think https://docs.vespa.ai/en/reference/page-result-format.html is a much better reference, would like to make it more like that (with the column widths unified). Yes on the examples, not sure if they should be in the reference or the guide, no strong opinion.

so, the json schema and examples will document the structure, and the descriptions of each field goes into a table is my suggestion

bratseth:
Sounds good to me. I think an example works best in the guide, the reference is for lookup and so needs a complete list/table of everything and not much else.

kkraune:
all right! yes that will make the guide better, too - will get to it. Thanks for bringing this up, @bartek !

kkraune:
reference and query-api guide improved in https://github.com/vespa-engine/documentation/pull/1853 - closing

PR #1836, opened 2022-02-15
Kkraune/fix examples

kkraune:
- note that the {% highlight xml%} liquid macro will correctly quote &lt; and the likes
- tested deployment locally


jobergum:
Great!

PR #1839, opened 2022-02-16
Place deprecation in text, not list

kkraune:
- do not break the list format


bratseth (approved):
I thought we agreed to remove it but ok.

PR #1840, opened 2022-02-16
Correct partition spec

kkraune:


vekterli (approved):
:+1: as discussed offline, we should deprecate this configuration parameter (and auto-generate a reasonable default, as on Vespa cloud) since it doesn't have the semantics people will come to expect from the values they put in.

kkraune:
yes! CC @gjoranv 

I assume @vekterli will drive the deprecation / better default in a separate PR

PR #1853, opened 2022-02-17
Kkraune/query api result format

kkraune:
- This transforms the multilevel list to a flat table. This removes the confusion about the structure of the results
- Examples moved to query API guide, also added an example for a regular query result
- Users will hence get the structure from the guide, or json schema (unchanged)


bratseth (approved):
👍

bratseth:
linking to internal hash #trace that does not exist (line 1500)
14:41:35      <a href="reference/default-result-format.html#trace">query tracing</a>

PR #1871, opened 2022-02-23
Use new, slimmer syntax

kkraune:
- targetNumHits is now targetHits
- ToDo on connectivity/id annotations, I think latest update to http://localhost:4000/en/text-matching-ranking.html#connectedness might be wrong, Searcher not required, can do this in YQL?


kkraune:
@geirst confirmed the ToDo, will make another update for connectedness in this PR

kkraune:
@geirst OK if you review instead of @bratseth :-)

geirst:
The connectivity clarification looks good!

PR #1874, opened 2022-02-23
Mention removed dispatch tuning settings

bratseth:


gjoranv:
We crossed each other on this one. I suggest to merge #1875 instead.

PR #1887, opened 2022-03-01
Add note how to use recall parameter with multiple labels

jobergum:


kkraune (approved):
Great! Consider adding to / link to/from query-api-reference.html as well for this

PR #1888, opened 2022-03-01
Suggested new format for query guide

kkraune:
- I want to make the query guide more accessible, with live examples for easier experimentation
- Here using Vespa documentation search, but can be anything that is always up
- Queries are in clear text (no urlencoding), click button to expand a pre with a curl command. Can add a copy button to the pre with the curl for easier copy.
- Hence much easier to add query examples in HTML code
- This is a draft for your feedback - if OK, can remove the existing and replace with this / something nicer :-)

queries

CC aesthetics committee @bratseth @frodelu

### DIFF
--- a/en/query-language.html
+++ b/en/query-language.html
@@ -130,4 +130,80 @@ field my_map type map&lt;string, int&gt; {
   </tbody>
 </table>
 
+
+<h2 id="query-examples">Query examples</h2>
+<p>
+  <button class="is-small"><span class="d-icon is-small d-arrowhead-down"></span></button> to expand queries.
+</p>
+
+<p><a class="docsearch-x">select * from doc where true</a></p>
+<p>
+  Select all documents from the "doc" source.
+</p>
+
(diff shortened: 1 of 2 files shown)


bratseth:
Yes, agreed we should move the repeated curl stuff asap. Also because people will use Vespa CLI now instead of curl.

bratseth:
Not sure we need the "expand" or something like that - could just be one single example with curl somewhere (if at all), it's completely orthogonal to formulating YQL.

kkraune:
Yes, happy to let the curl stuff go - it is a good idea to just have one example at the end with how to take the clear text YQL string and curl using
````
curl --data-urlencode ...
````
That will make the doc cleaner, too

kkraune:
... together with a vespa-cli example as well, of course

jobergum:
Looks good, I'm not sure what it expands to? A curl command or a link?

kkraune:
A curl command. But I will move this to one example in a separate section. Thanks for feedback!

PR #1892, opened 2022-03-01
Remove YQL semicolons

bratseth:


en/grouping.html
```diff
@@ -1312,21 +1312,21 @@ <h2 id="pagination">Pagination</h2>
 reproduce the same result by passing the <em>this</em> continuation along the original select:
 </p>
 <pre>
-select (&hellip;) | [{ 'continuations':['BGAAABEBCA'] }]all(&hellip;);
+select (&hellip;) | { 'continuations':['BGAAABEBCA'] }all(&hellip;)
```
kkraune, line 1315: Thanks! I was not sure if this is parsed as the annotations, so was not sure whether this can be removed

kkraune (approved):
Great!

en/grouping.html
bratseth: The outer one can, the inner is an array :-)

PR #1902, opened 2022-03-04
Move search box to the right

bratseth:
This avoids the visual issue caused by the misalignment of the
search box and left navigation pane.


kkraune (approved):
👍

PR #1906, opened 2022-03-07
Kkraune/query error handling

kkraune:
We should maybe also add an example of error handling in Searcher code - something like this is correct? 
````
        Result res = execution.search(query);
        for (ErrorMessage err : query.errors()) {
            // handle error
        }
````


bratseth (approved):

kkraune:
I see, no need then, thanks!

PR #1909, opened 2022-03-09
Remove <clients> element that only contained deprecated load type config

vekterli:
@kkraune please review
@gjoranv @ean @bratseth FYI

Load types are deprecated, so don't advertise their existence. Clients should
prefer using default priorities for feed, as there is a fine internal balance
between external and internal operations and their mutual priorities.

### DIFF
--- a/en/reference/services.html
+++ b/en/reference/services.html
@@ -14,7 +14,6 @@ redirect_from:
   <a href="services-content.html">content   [version]</a>
   <a href="services-admin.html">admin     [version]</a>
   <a href="services-routing.html">routing   [version]</a>
-  <a href="#clients">clients   [version]</a>
 </pre>
 Nodes/hosts referred to in <em>services.xml</em> must be defined in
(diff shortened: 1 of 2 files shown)


kkraune (approved):
Great!

kkraune:
or just write in the release notes that it i not possible to set pri anymore, and remove the section referring to it. no one should do this

vekterli:
We expose priority as part of quite a few (currently non-deprecated) public Java DocumentAPI surfaces, so this close to Vespa 8 I'm tempted to just remove the documentation of the _command line_ flags (which is what's failing the link checker) and yank it out of the APIs on Vespa 9. Thoughts?

kkraune:
I don't think it matters that we are close, noone probably uses this anyway. less work to do now, particularly because it is not _my_ work ;-)

gjoranv:
Sounds to me like we can handle this on 8, so if there isn't a ticket already, please create one @vekterli.

PR #1913, opened 2022-03-09
Front page improvements

bratseth:
- Explain what this is
- Remove the ugly figure
- Show the relevant links directly and remove the rest
- Show links as simple list
- Make the page search oriented

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,58 +4,36 @@ skipheading: true
 index: false
 ---
 
-<div class="row m-b-40">
+<div class="row m-20">
   <div class="col">
-    <div class="flex">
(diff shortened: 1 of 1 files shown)


kkraune (approved):
👍

kkraune:
Thanks. imho, it would make things better if we served this site on vespa.ai/documentation, having one less site - something to consider for later

kkraune:
need to look into how it renders on mobile phones, on mine the text does not break so must right scroll. probably due to the large search input.

ok to make this a little smaller (can also set same text size as rest of page to fit more text there, it is larger now)? Alternatively, must add logic to set size based on screen size

PR #1915, opened 2022-03-10
Navigation improvements

bratseth:
- No all-caps
- Don't indent selected item and simply bold it
- Heading and layout on TOC


kkraune:
Great! Much better! Nit: I think it is "collapsible" (per my dictionary)

kkraune (approved):

bratseth:
Both are legal :-)

PR #1924, opened 2022-03-14
Remove deep links and cleanup

bratseth:


kkraune:
The sidebar is built from the same content as https://docs.vespa.ai/sitemap.html - all pages are linked from there.

This to meet a request for all content being reachable directly. This made the sidebar a bit clumsier, otoh everything is there.

I don't necessarily think everything should be in the sidebar, maybe we should instead add a field to be able to hide some link from sidebar in https://github.com/vespa-engine/documentation/blob/master/_data/sidebar.yml? or any other ideas how to both manage the full TOC and the sidebar?

bratseth:
Everything wasn't linked, but I agree it's useful to have more stuff in the TOC. I guess we can add an attribute to the data?

kkraune:
agreed! assuming most will be in sidebar, we could try with something like this. OK to merge this, as we can fix the missing after

      - page: Build and install Vespa
        url: /en/build-install-vespa.html
        sidebar_exclude: true

kkraune (approved):

PR #1927, opened 2022-03-15
Support exluding from sidebar MERGEOK

kkraune:
- there are probably more that should be excluded from the sidebar
- add back a few docs from https://github.com/vespa-engine/documentation/pull/1924 for TOC


bratseth (approved):
👍

PR #1936, opened 2022-03-16
Matching and ranking with multi-valued fields. 

jobergum:
testing changes:

- enable PR testing - the test URL of vespa cloud getting started should be moved as it will fail as secrets are not available for PR testing. 
- To actually test PR builds one needs to use local paths instead of pulling from what is published 

new document demonstrates searching in weightedset set strings. 

### DIFF
--- a/en/searching-multi-valued-fields.md
+++ b/en/searching-multi-valued-fields.md
@@ -14,33 +14,35 @@ When building a search application we need to think about
 
 During the matching or retriveal stage, Vespa calculates several matching rank features 
 which can influence the order of the documents retrieved. This process is called ranking. A
-single query request might retrieve documents into ranking phase by multiple different 
+single query request might retrieve documents using multiple different 
 retriveal strategies combined using logical disjunction (OR).
-
 
 ### Matching
 
-There is a lot of matching options we should think about when 
-designing both our document schema and the way we query the schema
+There is a lot of text matching options we should think about when 
+designing and mapping or document model to a Vespa document schema:
 
-- For string fields we need to decide if we want to use text style matching or database-style exact matching?
-- For free text string fields, we need to think about;
-[linguistic processing](linguistics.html) like [tokenization](linguistics.html#tokenization), 
+- For string fields we should think about using text style matching or database-style exact matching.
+- For string fields there are also several
+[linguistic processing](linguistics.html) options like [tokenization](linguistics.html#tokenization), 
 normalization and language dependent [stemming](linguistics.html#stemming).
-- String fields which shares the same match and linguistic processing settings can be combined 
-  schema [fieldset](reference/schema-reference.html#fieldset). 
-- Should query matching in the field impact ranking, or is it just a filter which should not impact ranking order?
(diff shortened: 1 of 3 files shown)


en/searching-multi-valued-fields.md
```diff
@@ -0,0 +1,603 @@
...
+normalization and language dependent [stemming](linguistics.html#stemming).
+- String fields which shares the same [match](reference/schema-reference.html#match) 
+and linguistic processing settings can be combined using [fieldsets](reference/schema-reference.html#fieldset). 
+
+At query time, we can take the user query and translate it into a valid Vespa query request which implements
+our matching and retrieval strategy over the designed document schema.
+
+### Ranking 
+
+The documents which matches the query and is retrieved by the query is scored using a ranking model.
```
kkraune, line 37:
```suggestion
The documents which match the query and are retrieved by the query are scored using a ranking model. 
```

kkraune: or easier rewrite to "The document ..."

en/searching-multi-valued-fields.md
```diff
@@ -0,0 +1,603 @@
...
+
+But what about `black sky`:
+
+<pre data-test="exec" data-test-assert-contains='"totalCount": 1'>
+$ vespa query 'yql=select * from photos where userQuery()' \
+ 'query=black sky' 'type=all' 'default-index=tags'
+</pre>
+
+Also matches our document. This is an example of cross-element matching. With weightedset
+usig `indexing:index` with `match:text` multi term queries match across elements.
```
kkraune, line 289:
```suggestion
using `indexing:index` with `match:text` multi term queries match across elements.
```

en/searching-multi-valued-fields.md
```diff
@@ -0,0 +1,603 @@
...
+to match the set element exactly. `clear` matches `clear sky` and `sky` will match `clear sky`.  
+
+But what about `black sky`:
+
+<pre data-test="exec" data-test-assert-contains='"totalCount": 1'>
+$ vespa query 'yql=select * from photos where userQuery()' \
+ 'query=black sky' 'type=all' 'default-index=tags'
+</pre>
+
+Also matches our document. This is an example of cross-element matching. With weightedset
```
kkraune, line 288: consider linking to https://docs.vespa.ai/en/reference/query-language-reference.html#sameelement here, with a short note?

kkraune (approved):
This is a great guide! We can simplify guides/ docs in other places linking to this one instead, too, in later PRs. Thanks!

en/searching-multi-valued-fields.md
jobergum: Problem is that sameElement does not offer any ranking features :/

kkraune: right. I was thinking of just the matching part, and this is a frequent support question. I'll look at the doc again and see if there is a better place to just mention this

jobergum: sameElement is great, but only supported for the complex array of struct and map, which per today, is only supported for attribute fields, and has no rank feature support.

PR #1938, opened 2022-03-18
Move text-search tutorials to vespa-cli

jobergum:
text-search intro is now OK - but the ML part we should put more work into and harmonize with the other msmarco-ranking document runs using GBDT. 


kkraune (approved):
Great!

PR #1966, opened 2022-03-30
Move description of grouping session cache to reference doc

bjorncs:
Also add small section about correctness vs performance


baldersheim (approved):

kkraune (commented):
👍

PR #1973, opened 2022-04-01
Remove doc for deprecated splitter feature

kkraune:


gjoranv (approved):
Please note that there are references to a non-existent document in the header:
```
- /documentation/splitterjoiner.html
- /en/splitterjoiner.html
```

kkraune:
> Please note that there are references to a non-existent document in the header:
> 
> ```
> - /documentation/splitterjoiner.html
> - /en/splitterjoiner.html
> ```

Right, these are redirects from old links. no need in having those, I'll remove - thanks for spotting

PR #2015, opened 2022-05-05
Add doc for matching settings that can be overridden per query request.

geirst:
@jobergum please review
@baldersheim FYI

Depends on https://github.com/vespa-engine/vespa/pull/22464


kkraune (approved):
👍

PR #2027, opened 2022-05-10
Add API example usage

kkraune:
- Some text rewrites. sorry for all the reindent ...
- The real change is adding curl examples - it has shown too difficult with the reference API examples. maybe we should just remove those


kkraune:
ping @vekterli

PR #2031, opened 2022-05-11
Document tests dir in app package reference MERGEOK

kkraune:
- I will also change the album-recommendation sample app + docs in a separate PR set


kkraune:
... or @hmusum

hmusum (approved):

PR #2032, opened 2022-05-11
Use new dir structure for sampleapp MERGEOK

kkraune:
... or @jobergum


kkraune:
... or @hmusum

jobergum:
>ERROR: One or more files failed: en/vespa-quick-start.html, en/operations/monitoring.html

kkraune:
yes, this change must be merged at the same time as the change in sampleapps, so PR run will fail

jobergum (approved):

PR #2038, opened 2022-05-12
Add metric example

kkraune:


en/content/content-nodes.html
```diff
@@ -538,6 +538,38 @@ <h2 id="metrics">Metrics</h2>
   </tr>
 </tbody>
 </table>
+<p>
+  Example, using the <a href="../vespa-quick-start.html">quickstart</a> -
+  find the distributor port (look for HTTP):
+</p>
+<pre>
+$ docker exec vespa /opt/vespa/bin/vespa-model-inspect service distributor
```
vekterli, line 546: Do we want to have `docker exec` as part of this?

vekterli (approved):
:+1:

en/content/content-nodes.html
kkraune: I think so here, as it says it is using the quickstart, which is docker based, but one never execs into the docker image in the quickstart

PR #2046, opened 2022-05-18
Link query execution descriptions

kkraune:
- Use same illustration in both places
- It seems to me that these two sections overlap a lot, maybe we should have only one, moving some from the query API doc into the sizing doc - what do you think, @jobergum  ?


jobergum:
It's unclear to me what you mean, but feel free to correct duplication

PR #2052, opened 2022-05-19
Bratseth/inputs constants

bratseth:
@kkraune or @jobergum please review


jobergum:
> Hence the expression: as greedy as a pig

https://www.youtube.com/watch?v=2xUynRdzzsM

We need to wait until an open release with the fixes for inputs scalars,  I see test failures in the doc testing for it.

en/proton.html
```diff
@@ -552,7 +552,7 @@ <h2 id="proton-maintenance-jobs">Proton maintenance jobs</h2>
   <tr>
     <th>CPU</th>
     <td>CPU similar to feeding.
-        Consumes capacity from the index write thread - hence has feeding impact</td>
+        Consumes capacity from the index write threadm, so has feeding impact</td>
```
kkraune, line 555:
```suggestion
        Consumes capacity from the index write thread, so has feeding impact</td>
```

bratseth:
Yes, that'll be in Monday's release.

Out hence incidence is now down to just 10x normal ...

kkraune (approved):
Great!

PR #2075, opened 2022-06-07
Minor visiting doc cleanup

vekterli:
@kkraune please review

Remove some unnecessary fluff from the text as well as any mentions of using priorities.


kkraune (approved):
Better!

PR #2091, opened 2022-06-09
Move to json format doc

kkraune:


en/reference/default-result-format.html
```diff
@@ -538,3 +538,162 @@ <h2 id="json-schema">JSON Schema</h2>
     }
 }
 {% endhighlight %}</pre>
+
+
+<h2 id="appendix-legacy-vespa-7-json-rendering">Appendix: Legacy Vespa 7 JSON rendering</h2>
+{% include note.html content="This is work in progress, Vespa 8 has not been released yet."%}
```
arnej27959, line 544: not true real soon now

PR #2093, opened 2022-06-10
Add draft live query for grouping examples

kkraune:
We agreed to make the doc more clickable / live, so some sample grouping data is added to doc search in https://github.com/vespa-cloud/vespa-documentation-search/pull/15 (must be merged with this), based on https://github.com/vespa-engine/sample-apps/tree/master/examples/part-purchases-demo (this one can be removed later)

The idea is to easily write sample queries in clear text and auto-generate the link to the source.

@ldalves I would like these clickable examples stand out, maybe like in https://docs.vespa.ai/en/getting-started-ranking.html - but I would like your input on how to make this look great and intuitive

I just added one such query to the grouping guide for now so you can see how it works / change before we do more

FYI @jobergum


bratseth (approved):
👍

PR #2097, opened 2022-06-14
Revert back to gzip as default for access log compression

gjoranv:


arnej27959:
default change in progress
https://github.com/vespa-engine/vespa/pull/23102

PR #2103, opened 2022-06-20
Document what time in access log means

hmusum:
Point to detailed explanation in reference doc

@bjorncs, @baldersheim @bratseth might know if this is correct and/or if we need more explanation


bjorncs (approved):

bjorncs (changes requested):
It's the time when the request is received by jdisc

bjorncs (approved):

PR #2104, opened 2022-06-20
Document max field size for bolding

kkraune:
- ref https://github.com/vespa-engine/vespa/pull/23168
- Merge this after Vespa 8 is available


geirst (approved):
👍

PR #2108, opened 2022-06-21
Complete YQL changes

kkraune:
- For completeness, teams must update unit tests in some cases
- I suppose we should deprecate the old format for removal in Vespa 9 right away?


bratseth (approved):

bratseth:
> I suppose we should deprecate the old format for removal in Vespa 9 right away?

I feel it is better to wait so we don't reward those who just did the work of removing deprecations with a new set of them. And the cost of this is very low so I'm fine with keeping them until 2029 or so.

PR #2109, opened 2022-06-21
Kkraune/doc v1 ref

kkraune:
- The examples are also in the guide, so remove from here
- Table format cleanup to align columns
- Split into mandatory and optional parameters, add missing

Only need to review the last commit, the others are just moving text around / format changes / deleting the obvious. Left ToDo for some parameters I am not sure about, please comment

### DIFF
--- a/en/reference/document-v1-api-reference.html
+++ b/en/reference/document-v1-api-reference.html
@@ -40,19 +40,19 @@ redirect_from:
       <tr>
         <th>Get</th>
         <td>
-          Get a single document:
+          Get a document:
 <pre>
 /document/v1/&lt;namespace&gt;/&lt;document-type&gt;/docid/&lt;document-id&gt;
-</pre>
-          Get a document, using grouped numeric ID (see below):
-<pre>
 /document/v1/&lt;namespace&gt;/&lt;document-type&gt;/number/&lt;numeric-group-id&gt;/&lt;document-id&gt;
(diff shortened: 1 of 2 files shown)


en/reference/document-v1-api-reference.html
```diff
@@ -30,290 +30,270 @@ <h2 id="configuration">Configuration</h2>
...
+          Note that visit via selection is evaluated over all the stored documents in the cluster.
+          To split visiting of the document corpus across more than one HTTP
+          request—thus allowing the concurrent use of more HTTP containers—use the
+          <a href="#slices">slices</a> and <a href="#sliceId">sliceId</a> parameters.
+        </li>
+        <li>
+          It's also recommended enabling streamed HTTP responses,
+          with the <a href="#stream">stream</a> parameter,
+          as this reduces memory consumption and reduces HTTP overhead.
+          Finally, use the <a href="#concurrency">concurrency</a> parameter
```
jonmv: Only true when `stream` is _not_ used. When `stream=true`, concurrency _limits_ the maximum concurrency, which is otherwise unbounded, but controlled by a dynamic throttle policy.

kkraune:
ptal @jonmv

jonmv:
I don't think the stuff you removed from the reference actually got moved to the guide? Otherwise LGTM.

PR #2124, opened 2022-06-24
Remove vespa7 text from reference

kkraune:
- linking to the geo doc will suffice
- I am not sure if the Vespa 7 `latlong` description should be anywhere, @arnej27959 ?

### DIFF
--- a/en/geo-search.html
+++ b/en/geo-search.html
@@ -32,13 +32,10 @@ redirect_from:
   east (longitudes); negative numbers are used for south and west.
   This is the usual convention.
 </p>
-<p>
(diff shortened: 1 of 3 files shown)


arnej27959 (approved):
looks good but now reference from
https://docs.vespa.ai/en/reference/query-api-reference.html#pos.ll
needs to be removed. (mark pos.* deprecated, say "use YQL ...")

kkraune:
Thanks! There is a deprecation warning just above for all the pos.* parameters - it seems they are still in Vespa 8, deprecated for Vespa 9 I assume?

PR #2128, opened 2022-06-25
Document restructured tracing parameters

bratseth:


kkraune:
The geo changes intentional?

bratseth:
What geo changes?

kkraune:
how odd. I was reviewing using the GH mobile app, but must have been user error (no idea where the other changes came from). This looks great!

kkraune (approved):
one link that must be updated, see the link check PR run

PR #2131, opened 2022-06-27
Document Summaries are memory-only operations

bjormel:


bjormel:
@kraune, maybe we can remove "Use attributes in document summaries to limit accesses to storage to generate result sets.", now that we are listing this a one of the "key" features of attributes

kkraune:
agreed

kkraune (approved):

PR #2139, opened 2022-07-01
Kkraune/geo cleanup

kkraune:
- note: pos.attribute is not mentioned in vespa8-release-notes.html, this is an oversight? This PR removes it from doc as it is the only geo parameter left in the query API

### DIFF
--- a/en/reference/query-api-reference.html
+++ b/en/reference/query-api-reference.html
@@ -112,15 +112,6 @@ get methods as Java objects from the Query to Searcher components.
   </ul>
 </dd>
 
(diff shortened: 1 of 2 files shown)


arnej27959 (approved):
👍

kkraune:
OK - so I should add removal of pos.attribute to vespa 8 release notes as well, @arnej27959 ?

arnej27959:
> add removal of pos.attribute to vespa 8 release 

this parameter only made sense when using pos.ll also, but for completeness you can add both pos.attribute and pos.bb if you want.

PR #2141, opened 2022-07-01
Add documentation for specifying metric sets

frodelu:
Attempt for documenting my understanding of this, but would like to have both @gjoranv and @olaaun review of accuracy. Also need to decide if it is OK to "expose" the `vespa` metric-set and if the disclaimer that this is "stable" is OK, @bratseth .

FYI: @kkraune

### DIFF
--- a/en/reference/metrics.html
+++ b/en/reference/metrics.html
@@ -35,8 +35,15 @@ These apps also include examples for how to find ports used by using
 <a href='vespa-cmdline-tools.html#vespa-model-inspect'>vespa-model-inspect</a>."%}
 
 
+<h2 id="aggregated-metrics">Aggregated metrics</h2>
+<p>
+    For most cases, metrics should be collected aggregated on an application-level through the
+    <em><a href="#metrics-v2-values">/metrics/v2/values</a></em> API or at node-level through either the
+    <em><a href="#metrics-v1-values">/metrics/v1/values</a></em> or
+    <em><a href="#prometheus-v1-values">/prometheus/v1/values</a></em> API.
+</p>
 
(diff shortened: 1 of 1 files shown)


en/reference/metrics.html
```diff
@@ -172,9 +179,72 @@ <h2 id="metrics-v1-values">/metrics/v1/values</h2>
...
+            <td>
+                The default <a href="https://github.com/DataDog/integrations-extras/blob/master/vespa/metadata.csv">set of metrics</a>
+            </td>
+        </tr>
+        <tr>
+            <td>vespa</td>
+            <td>
+                A <a href="https://github.com/vespa-engine/vespa/blob/master/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java">comprehensive list of metrics</a>
+                used to inspect low-level behavior of Vespa.<br/>
+                <b>NOTE:</b>The set of metrics here are not guaranteed to stay stable within a minor version of Vespa.
```
olaaun, line 216: We only remove metrics from this set on new major versions, no?

bratseth: yes

bratseth:
Yes, the "vespa" metric set is ok to expose.

en/reference/metrics.html
frodelu: Should we update text to remove note or leave it as-is?

bratseth: remove

gjoranv: The note could instead warn about the potential cost of sending large number of metrics to external systems?

gjoranv (approved):
👍

PR #2144, opened 2022-07-05
Clarify/extend some points in our consistency documentation

vekterli:
@bratseth please review

### DIFF
--- a/en/reads-and-writes.html
+++ b/en/reads-and-writes.html
@@ -31,7 +31,7 @@ redirect_from:
   and multiple container clusters - see <a href="indexing.html">indexing</a> for details.
 </li>
 <li>
-  Writes in Vespa are <em>eventually consistent</em>, but Vespa will prioritize availability over consistency when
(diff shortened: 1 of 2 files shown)


en/reads-and-writes.html
```diff
@@ -31,7 +31,7 @@
   and multiple container clusters - see <a href="indexing.html">indexing</a> for details.
 </li>
 <li>
-  Writes in Vespa are <em>consistent</em>, but Vespa will prioritize availability over consistency when
+  Writes in Vespa are <em>eventually consistent</em>, but Vespa will prioritize availability over consistency when
```
bratseth, line 34: How do you think it's "eventually"? Usually this means "after a write I may get the old or the new value, but eventually it will converge to the new".

bratseth: > Maybe it would be better to rephrase to state that in a stable system, writes are consistent, but otherwise we will prioritize availability?

Yes - I think that is better. That is our design goal and people usually get it when explained in this way.

vekterli: @bratseth I've changed the wording here, PTAL

PR #2151, opened 2022-07-12
Update live upgrade procedure now that detaching of nodes is not nece…

hmusum:
…ssary anymore.


bratseth (approved):
We should wait a bit with merging though.

hmusum:
Added section on Vespa 7 upgrade to release notes, changed live upgrade doc to point to this section for vespa 7 upgrades

hmusum:
@bratseth and/or @kkraune please take a look

bratseth (approved):
💯

PR #2168, opened 2022-07-26
Kkraune/restart

kkraune:
- document the stop/start commands in the reference, and link to these
- generally, make it easier to use the correct vespa-stop-services command for node restart


arnej27959 (approved):
👍

PR #2185, opened 2022-08-03
Kkraune/faq

kkraune:
- I did not find a good way to write this in the searcher doc, pasted into a FAQ entry instead for now


bratseth (approved):
👍

PR #2188, opened 2022-08-04
Workaround: fix list by removing a blank line

kkraune:
- if it is important to keep the blank line after the import statement, we can work around using html most likely, but this works


thigm85 (approved):
LGTM

PR #2192, opened 2022-08-10
Kkraune/singlenode

kkraune:


en/operations/multinode-systems.html
```diff
@@ -316,6 +317,64 @@ <h3 id="vespa-nodes-setup">Vespa nodes setup</h3>
...
+  <li>
+    Install Vespa:
+<pre>
+$ sudo dnf config-manager \
+  --add-repo https://raw.githubusercontent.com/vespa-engine/vespa/master/dist/vespa-engine.repo
+$ sudo dnf config-manager --enable powertools
+$ sudo dnf install -y epel-release
+$ sudo dnf install -y vespa
+$ export VESPA_HOME=/opt/vespa
+$ echo "override VESPA_CONFIGSERVERS ip-172-31-95-248.ec2.internal" | \
```
ean, line 341: I expect this will result in users copying it verbatim and needing to debug why this failed.
does `hostname` resolve to the same host. If so it can be replaced with `$(hostname)` and there is one less thing that may go wrong.

kkraune: I agree. otoh, this setting is often set wrong in a multi-config-server setup, so it is better that users get used to setting actual values here - so I wanted to keep it the same as the multinode setup above, but using only one name, to highlight the difference.

ean (approved):

en/operations/multinode-systems.html
ean: sounds good

PR #2193, opened 2022-08-10
Add namespace example / FAQ entry

kkraune:
- adding great slack response as an example
- add namespace in heading for better doc search ranking


jonmv (approved):
Hehe :)

PR #2198, opened 2022-08-10
Define parametrized pre-reqs

kkraune:
- All guides will have the same set of pre-requisites, possibly with different memory settings
- This gives a common look and feel and is easy to update in one place
- The quickstart runs well on 4G (was 6G)

FYI @jobergum 

ref https://github.com/vespa-engine/vespa/issues/20788

Skjermbilde 2022-08-10 kl  15 57 50


jobergum:
Are you able to to this for the markdown readmes of the sample apps?

kkraune:
Not easily, @jobergum : "You can’t put <style> or <script> tags in it". But I will go through the readme files and put the same block in all (with correct memory and other settings after testing) for the same look and feel. So it will be better.

jobergum (approved):

PR #2199, opened 2022-08-11
Mention the safe startup option

kkraune:
@arnej27959 : the docs says "The safe mode is disabled by default" - but it seems is is default on for self-hosted Vespa, ref https://github.com/vespa-engine/sample-apps/tree/master/examples/operations/multinode-HA#start-the-admin-server ?


kkraune:
ping @arnej27959

arnej27959:
Yes - we decided to make enabled by default after seeing that it works as intended. So wording here needs to be updated.

kkraune:
thanks, will fix wording!

kkraune:
@arnej27959 PTAL

jobergum (approved):

PR #2224, opened 2022-08-24
Correct the weight in the explanation in news 3 tutorial

anthager:
In the last part of the [news 3 tutorial](https://docs.vespa.ai/en/tutorials/news-3-searching.html), the weight in the explanation is not the same as in the code above (100 in the explanation and 10 in the code). See screenshot. The explanation is therefore changed to 10.

Screenshot 2022-08-24 at 10 28 18


bratseth (approved):
👍

PR #2235, opened 2022-08-26
Move example to the guide, adapt to quickstart

kkraune:
- This is work in progress and not complete at all, but get something in that is not wrong for now
- some unrelated reindentation while at it in the attribute doc
- 
@geirst or @baldersheim

### DIFF
--- a/en/phased-ranking.html
+++ b/en/phased-ranking.html
@@ -159,7 +159,8 @@ by not evaluating the real ranking function on all the hits.
   run a query and order by the mutable attribute.
 </p>
 {% include note.html content='The mutable attributes are just counters and memory-operations only -
-the values do not survive content node restarts. The values cannot be compared across nodes.
+the values might or might not not survive content node restarts.
+The values cannot be compared across nodes.
 Use the values to assess relative document matching and ranking performance since Vespa start' %}
 <p>
   This example is based on the <a href="vespa-quick-start.html">quickstart</a>.
@@ -229,12 +230,9 @@ schema music {
         }
     }
 
-    rank-profile rank_albums_reset inherits rank_albums {
+    rank-profile rank_albums_reset_on_match inherits rank_albums {
         mutate {
             on-match        { match_count         = 0 }
-            on-first-phase  { first_phase_count   = 0 }
-            on-second-phase { second_phase_count  = 0 }
-            on-summary      { summary_count       = 0 }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


en/phased-ranking.html
```diff
@@ -126,3 +126,180 @@ <h2 id="choosing-phased-ranking-functions">Choosing phased ranking functions</h2
...
+  <li>on-summary</li>
+</ol>
+<p>
+  The common use case is to increase the value by 1 for each execution.
+  With this, it is easy to evaluate the document's performance to the queries,
+  e.g. find the documents that appear in most queries, or the ones that never matched -
+  run a query and order by the mutable attribute.
+</p>
+{% include note.html content='The mutable attributes are just counters and memory-operations only -
+the values do not survive content node restarts. The values cannot be compared across nodes.
```
geirst, line 162: Note: They might survive restarts if the mutable attribute was flushed to disk in the meantime, but this is something you cannot rely on.

en/phased-ranking.html
```diff
@@ -126,3 +126,180 @@ <h2 id="choosing-phased-ranking-functions">Choosing phased ranking functions</h2
...
+    indexing: 7 | to_long | attribute | summary   # Initialized to 7 for a new document. Default is 0.
+    attribute: mutable
+}
+</pre>
+<p>
+  To dump values fast, from memory only (assuming the schema has an <code>id</code> field):
+</p>
+<pre>
+document-summary rank_phase_statistics {
+    summary id type int { source: id }
```
geirst, line 294: I think the _source_ can be skipped for these five fields.

en/phased-ranking.html
```diff
@@ -126,3 +126,180 @@ <h2 id="choosing-phased-ranking-functions">Choosing phased ranking functions</h2
...
+            on-match        { match_count        += 1 }
+            on-first-phase  { first_phase_count  += 1 }
+            on-second-phase { second_phase_count += 1 }
+            on-summary      { summary_count      += 1 }
+        }
+    }
+
+    rank-profile rank_albums_reset inherits rank_albums {
+        mutate {
+            on-match        { match_count         = 0 }
```
geirst, line 234: To get proper reseting of all 4 mutable attributes, you must define 4 rank profiles where _on-match_ sets the value for that attribute to 0. Only _on-match_ is guaranteed to be executed for all documents when using a 'true' query.

geirst (approved):
👍

en/phased-ranking.html
kkraune: Good point! will redo this part and add a ToDo for the rest

PR #2245, opened 2022-08-30
Document /state/v1 API

kkraune:
- Add a proper reference for /state/v1
- Move /state/v1 guide text from reference/metrics.html to operations/monitoring.html. It could be these guide pieces should be elsewhere, but fits quite well, at least better than before
- reference/metrics.html must still be reworked to split out more ref doc for other APIs like metrics/v* and the likes - that is for a separate PR
- Need to work out a better structure in the sidebar later, maybe a separate API reference section?
- left ToDos for missing items, will consult @gjoranv  and @hmusum - this work is not fully completed, but the structure is right now


bratseth (approved):
Great!

PR #2247, opened 2022-08-31
Remove initializing state

kkraune:
- all this text would benefit from a rewrite / compaction, on the list for later


vekterli (approved):
:+1:

PR #2251, opened 2022-09-02
Kkraune/metrics reference

kkraune:
- metrics API reference doc is now clean. It is missing request params, though, at least `consumer`, but there are maybe more @gjoranv / @hmusum ? That is for next PR
- moved the residual doc into a metrics guide, that I will refine / clean up next

FYI @bratseth


bratseth:
👍

hmusum (approved):

gjoranv:
🎖️

PR #2253, opened 2022-09-05
remove cloud-only doc link

kkraune:
- does not belong here


bratseth (commented):
"Vespa can be deployed in multiple ways. These guides show how to deploy multi-node applications in various environments."

Vespa Cloud is an environment in which this can be done, so I don't see why not?

hmusum (approved):

PR #2255, opened 2022-09-06
Reformat - like the other API references

kkraune:
- no changes to content
- however: the current doc is inconsistent - is `-` valid in tenant names or not? the doc says yes and no ... lets clean up thsi before merging


hmusum:
Tenant name must match this pattern:

`static final Pattern namePattern = Pattern.compile("(?!\\.\\.)[a-zA-Z0-9_-]{1,256}");`

(from https://github.com/vespa-engine/vespa/blob/master/config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java#L19), so `-` is allowed.

kkraune:
@hmusum PTAL

hmusum (approved):

PR #2264, opened 2022-09-13
Kkraune/quote

kkraune:
- include some XML content in search
- un-quote some characters before feeding

approve only, please


kkraune:
... or @hmusum please approve (will roll back as needed)

hmusum (approved):

PR #2265, opened 2022-09-13
Experiment with links in headings

kkraune:
- added to https://docs.vespa.ai/en/overview.html so we can dry-run it there
- this adds a link in headers when hovering over, a la readme-files on GitHub and others

... or @frodelu

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -254,6 +254,10 @@ h1[id]::before, h2[id]::before, h3[id]::before, h4[id]::before, p[id]::before {
         margin-top: 20px;
         margin-bottom: 10px;
     }
+
+    h2:hover > a {
+        opacity: 1;
+    }
 }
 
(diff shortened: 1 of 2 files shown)


frodelu (approved):
As a Javascript-way this would probably work so seems OK. Ideally, I had hoped we could do this process time, but I haven't found a smooth way to do that with Jekyll yet.

frodelu:
Leaving it up to you to merge when ready, @kkraune …

kkraune:
Agreed, it would be better to add this when building the site. some Ruby processing / liquid, will think of it

kkraune:
note to self: looks ok, but add check for empty id, don't add then. we could auto-generate the id if not already set, too, but that makes auto link checking harder, so probably not (another argument for doing this at site build time). otoh, setting manually will resolve links in IDE when writing -> keep manual.

PR #2283, opened 2022-09-20
SEO

bratseth:
Maybe we should add a keywords field to the doc schema and rank matches high?


kkraune (approved):
keywords: Great idea, I'll get to it today

PR #2295, opened 2022-09-30
Kkraune/linguistics

kkraune:
- which one is correct for query language setting? https://docs.vespa.ai/en/reference/query-api-reference.html#model.locale or https://docs.vespa.ai/en/reference/query-api-reference.html#model.language
- This PR adds some details on language detection - this will be integrated in query recall troubleshooting

### DIFF
--- a/en/linguistics.html
+++ b/en/linguistics.html
@@ -17,6 +17,42 @@ redirect_from:
   <li>finding the base form of words (stemming or lemmatization).</li>
 </ul>
 <p>
+  Linguistic processing is run when writing documents, and when querying:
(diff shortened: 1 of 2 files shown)


bratseth (approved):
Great!

geirst:
@kkraune : Seems like the new illustration is missing. From https://docs.vespa.ai/en/linguistics.html:
Screenshot 2022-10-03 at 11 26 27

kkraune:
thanks! added in https://github.com/vespa-engine/documentation/pull/2296

PR #2333, opened 2022-10-12
Document possibiliity of visiting at root level

jonmv:
@vekterli please review.

### DIFF
--- a/en/reference/document-v1-api-reference.html
+++ b/en/reference/document-v1-api-reference.html
@@ -80,8 +80,12 @@ redirect_from:
       <p id="visit">
         Iterate over and get all documents, or a <a href="#selection">selection</a> of documents,
         in chunks, using <a href="#continuation">continuation</a> tokens to track progress.
-        Visits is implemented as a linear scan over the documents in the cluster.
+        Visits are a linear scan over the documents in the cluster.
       </p>
+<pre>
+/document/v1/
+</pre>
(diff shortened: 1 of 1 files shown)


en/reference/document-v1-api-reference.html
```diff
@@ -80,9 +80,13 @@ <h2 id="http-requests">HTTP requests</h2>
       <p id="visit">
         Iterate over and get all documents, or a <a href="#selection">selection</a> of documents,
         in chunks, using <a href="#continuation">continuation</a> tokens to track progress.
-        Visits is implemented as a linear scan over the documents in the cluster.
+        Visits are a linear scan over the documents in the cluster.
       </p>
 <pre>
+/document/v1/
```
vekterli, line 86: Technically requires a `cluster` parameter, so just substituting this in verbatim from the docs will fail. Include here as well?

vekterli (approved):
:+1:

en/reference/document-v1-api-reference.html
jonmv: I considered that, but it's not what is done elsewhere on this page. Perhaps we should do it _all the places_, then?

jonmv: Hmm ... that makes for some very long lines. 
Since this isn't a guide, and you can't copy working examples from it, and also because the mandatory parameters are listed just a few lines further down, I think it's fine as it is?

PR #2334, opened 2022-10-13
Document vespa-analyze-onnx-model

kkraune:
FYI @lesters @jobergum


jobergum:
Does the example need to dig the data out of the configuration server store?

kkraune:
not really, but the tool exists in the running container, and you then have to map the model file into that container to be able to analyze it. maybe there is an easier way

havardpe (approved):

PR #2341, opened 2022-10-17
Add link to stack overflow query

kkraune:
- I support we should co-locate the un(escape) functions in one module, I struggled with the export/import so leaving this for now

### DIFF
--- a/js/handle_results.js
+++ b/js/handle_results.js
@@ -1,6 +1,25 @@
-const handleResults = (data) => {
+
+const unescapeMap = Object.freeze({
+  '&amp;' : '&',
+  '&lt;'  : '<',
+  '&gt;'  : '>',
+  '&quot;': '"',
+  '&#39;' : "'",
+  '&#x2F;': '/',
+  '&#x60;': '`',
+  '&#x3D;': '='
+});
+const unescapeHtml = (string) => String(string)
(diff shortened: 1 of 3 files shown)


js/handle_results.js
```diff
@@ -1,7 +1,26 @@
...
+    .replace(/&amp;|&lt;|&gt;|&quot;|&#39;|&#x2F;|&#x60;|&#x3D;/g, (s) => unescapeMap[s]);
+
+
+const handleResults = (data, query) => {
   const result = document.getElementById("result");
   result.innerHTML = "";
 
+  let a = document.getElementById("stackoverflow-query")
+  a.innerHTML = "[vespa] " + query;
+  a.href = encodeURI("https://stackoverflow.com/search?q=[vespa] " + unescapeHtml(query));
```
freva, line 22: This should use `URLSearchParam` to properly encode the query. `encodeURI()` does not escape special URI characters, e.g. `&`. Maybe something like
```
const searchParams = new URLSearchParams();
searchParams.set('s', '[vespa] ' + unescapeHtml(query));
a.href = 'https://stackoverflow.com/search?' + searchParams.toString();

kkraune: agree, that is better, fixing

freva (approved):

PR #2356, opened 2022-10-21
Add FAQ entry for memory usage

kkraune:

CC @jobergum


bratseth:
I think the main confusion people have is between different metrics for measuring memory consumption, this entry would be greatly improved by talking about that.

kkraune:
Agree! I'll improve it

geirst (commented):
👍

PR #2357, opened 2022-10-21
Use inputs

kkraune:
- use input and remove query profile entry
- this fixes an auto query test
- Its in my plan to instead just extract query links and match to a regex in a screwdriver.cd instead of the current clunky test annotation and duplicated code - motivation increased now

There are a few more changes needed for perfection, next PR


bratseth (approved):
👍

PR #2360, opened 2022-10-25
Use 8 MERGEOK

bratseth:


kkraune (approved):
Thanks! Can also use {{site.variables.vespa_version}} - this expands to the value in _config.yml, this is used elsewhere in doc. I will auto-update this variable in an SD job very soon

bratseth:
That is much better - thanks

PR #2365, opened 2022-10-25
Add drop schema procedure

kkraune:
- review only - I need the latest vespa version to confirm it works with `schema-removal` - this was recently added as an alias

### DIFF
--- a/en/operations/batch-delete.html
+++ b/en/operations/batch-delete.html
@@ -46,4 +46,69 @@ $ curl -X DELETE --cert data-plane-public-cert.pem --key data-plane-private-key.
   "$ENDPOINT/document/v1/my_namespace/my_doctype/docid?selection=true&cluster=my_cluster"
 {% endhighlight %}</pre>
   </li>
+  <li>
+    <p id="drop-schema">
+      It is possible to drop a schema, with all its content, by removing the mapping to the content cluster.
+      To understand what is happening, here is the status before the procedure:
+    </p>
+<pre>
+# ls $VESPA_HOME/var/db/vespa/search/cluster.music/n0/documents
+
+drwxr-xr-x 6 vespa vespa 4096 Oct 25 16:59 books
+drwxr-xr-x 6 vespa vespa 4096 Oct 25 12:47 music
+</pre>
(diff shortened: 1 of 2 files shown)


kkraune:
tested with 8.73, this is good to go

en/operations/batch-delete.html
```diff
@@ -46,4 +46,69 @@
...
+<pre>
+# ls $VESPA_HOME/var/db/vespa/search/cluster.music/n0/documents
+
+drwxr-xr-x 6 vespa vespa 4096 Oct 25 17:06 books
+drwxr-xr-x 6 vespa vespa 4096 Oct 25 12:47 music
+</pre>
+    <p>
+      The procedure, deploying with and without the schema, is an efficient way to drop all documents.
+      It is good practise to remove <em>validation-overrides.xml</em>
+      or the <code>content-type-removal</code> element inside, to avoid subsequent accidental data loss.
```
geirst, line 101:
```suggestion
      or the <code>schema-removal</code> element inside, to avoid subsequent accidental data loss.
```

en/operations/batch-delete.html
```diff
@@ -46,4 +46,69 @@
...
+      The directory listing is just for illustration.
+    </p>
+    <p>
+      Use <a href="../reference/vespa-cmdline-tools.html#vespa-visit">vespa-visit</a> to list documents:
```
geirst, line 105: Is this last section required?

geirst (approved):
👍

en/operations/batch-delete.html
kkraune: not really - I included it as a way for people to easily verify it worked. will consider removing / linking elsewhere

geirst: I am also concerned that running this command on a system with many documents will take a long time. Then it is better to use vespa-proton-cmd (https://docs.vespa.ai/en/reference/vespa-cmdline-tools.html#vespa-proton-cmd) with either getProtonStatus or getState to verify. I think one of those returns details on document counts.

kkraune: good point - will update - thanks!

PR #2380, opened 2022-11-01
Kkraune/todos MERGEOK

kkraune:
- Clean up the hist/offset and maxHist/maxOffset documentation

Bonus: it seems `sorting.degrading` is not documented in the query api reference, will add it in a separate PR


geirst (approved):
👍

PR #2382, opened 2022-11-01
Add reference doc for sorting.degrading

kkraune:
FYI @bratseth I guess this is an accidental omission from the ref doc, adding in the "Other section for now"


kkraune:
FYI @freva - this is a parameter that was not documented - consider adding to the query builder for completeness

geirst (approved):
👍

freva:
Added to query builder in #24705

PR #2388, opened 2022-11-03
Multi-threaded pre-filter evaluation with ANN is supported in 8.78.45.

geirst:
@jobergum please review


kkraune (approved):
Great! Adding to next newsletter, you can think of a few lines for that, @geirst  ;-)

PR #2392, opened 2022-11-04
First cut vector search landing page

kkraune:
- Not linked from anywhere yet, there are probably major things missing / links that should go elsewhere


en/vector-search.html
```diff
@@ -0,0 +1,248 @@
...
+  in the context of the current song or an image being viewed.
+  This is a problem found in recommender systems.
+</p>
+
+
+
+<h2 id="digital-representations">Digital representations</h2>
+<p>
+  A solution to the synonym or multi-modal problem is to change from matching in the textual to the digital domain.
+  This means transforming text, images, and songs into a set of numbers that describe their <em>meaning</em> of it.
```
jobergum, line 69: I think that it's a stretch to use the word meaning here.

jobergum (commented):
Great, please consider review comments. 

I think a section on choosing vectorization models is missing.

jobergum:
I think we can iterate on this, but I think we should have

- A section on the shortcomings when using vector search alone using off-the-shelf models, but that hybrid search helps on this. Link to the two talks on this from BB. 
- A section on choosing vectorization model and how to use embeddings made easy, how to convert etc.

en/vector-search.html
```diff
@@ -0,0 +1,248 @@
...
+<p>
+  Note that ANNs require some kind of indexing to speed up search,
+  so inserts (adding a new vector) are more expensive (uses more CPU) and takes more space (memory and disk).
+  When evaluating different kinds of ANN indexing,
+  consider if your use case requires updates, including deletes, to the vectors - Vespa supports all.
+</p>
+
+
+
+<h2 id="optimizing-vector-search">Optimizing vector search</h2>
```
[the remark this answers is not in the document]
kkraune: ok, comment out and left a todo

en/vector-search.html
kkraune: all right, softened the language to indicate what it is about

kkraune:
> I think we can iterate on this, but I think we should have

good points, thanks for the great review! Should we address the above first, or merge and add soon?

PR #2393, opened 2022-11-07
Kkraune/timeout

kkraune:


kkraune:
ping @baldersheim - this moved the timeout doc to the reference, a better fit for it

kkraune:
@geirst this is moving the text

geirst (approved):
👍

PR #2407, opened 2022-11-10
let quickstart point to langauge issues

kkraune:
- stemming and language is important - let first-time users try this


bratseth (commented):
This always happens to the quick starts: Issue of the day seems important enough to put in the quick start and then after a while it is pages and pages long.

... so I think we should keep it minimal. _Maybe_, at most, add a list of links.

kkraune:
@bratseth PTAL

bratseth (approved):

PR #2443, opened 2022-12-05
Link to exact match alternative + query example

kkraune:
- it is easy to get the encoding wrong when testing ...

### DIFF
--- a/en/query-rewriting.html
+++ b/en/query-rewriting.html
@@ -386,7 +386,7 @@ consider using <a href="reference/schema-reference.html#match">match: exact</a>,
 </pre>'%}
 <p>
   Remember to encode queries when testing with non-textual characters.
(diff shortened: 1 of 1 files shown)


kkraune:
maybe I missed your point - I made this for cut and paste from the generated YQL that I think you can just use - or run the curl command

bratseth:
But if you copy and paste into the command line it'll be wrong?

bratseth:
`vespa query -v 'select * from items where product contains "M&amp;M"`

is copied as "curl etc."?

kkraune:
yql

I'm not sure I understand the problem - when you copy from the page the text copies as is - both the vespa cli command and the url

bratseth:
Yes, right, no problem. :-)

bratseth (approved):

PR #2445, opened 2022-12-06
Kkraune/http2 MERGEOK

kkraune:
- from the slack thread last month - just review the last commit

Note: I also see 1.1 for GET:
```
$ curl -i http://127.0.0.1:8080/search/\?timeout=10s\&yql=select+%2A+from+music+where+true
HTTP/1.1 200 OK

$ curl -i --http2-prior-knowledge http://127.0.0.1:8080/search/\?timeout=10s\&yql=select+%2A+from+music+where+true
HTTP/2 200 
```

### DIFF
--- a/en/performance/http2.html
+++ b/en/performance/http2.html
@@ -4,92 +4,113 @@ title: "HTTP/2"
 redirect_from:
 - /documentation/performance/http2.html
 ---
+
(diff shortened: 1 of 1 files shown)


bjorncs (approved):
👍

PR #2455, opened 2022-12-09
Update procedure for cfg server scaling

kkraune:


en/operations/configuration-server.html
```diff
@@ -206,64 +206,57 @@ <h2 id="start-sequence">Start sequence</h2>
...
+  Read up on <a href="#zookeeper-configuration">ZooKeeper configuration</a> before continuing.
+  Although it is <em>possible</em> to add more than one config server at a time,
+  doing it one by one is recommended, to keep the ZooKeeper quorum intact.
+</p>
+<p>
+  Due to the ZooKeeper majority vote, use one or three config servers.
+</p>
+<p>
+  To <em>replace</em> config servers, the easiest way is to remove a server and add the new one with the same name -
+  updating <em>VESPA_CONFIGSERVERS</em> is then not needed, and no nodes need restarting.
```
hmusum, line 219: I am unsure if what we are saying about no nodes needing restart is correct. Clients will need to lookup IP from hostnames often/regularly for this to work. Then again, I think we do this in hosted Vespa, so...

kkraune: right. so not sure what do put here

hmusum: I'll pick this up again over the weekend and discuss this with some others to see what makes sense

hmusum:
I have merged master into this branch and done some minimal changes while doing so

kkraune:
we should decide whether to merge or close, @hmusum

hmusum:
Should have been tested in some way, but I don't know how to easily do that. I still think that this is an improvement, so will merge.

PR #2459, opened 2022-12-13
Fix typo in scaling law documentation page

Hadrien-Cornier:
Amdahl's law is 1/(1-DQW/(DQW+SQW)) and not 1/(1-SQW/(DQW+SQW)) but DQW is match_time and SQW is query_setup_time

### DIFF
--- a/en/performance/sizing-search.html
+++ b/en/performance/sizing-search.html
@@ -642,7 +642,7 @@ By sampling these metrics,
 one can calculate the theoretical speedup we can achieve by increasing number of nodes using flat distribution,
 by using <em>Amdahl's law</em>:
 </p><p><!-- depends on mathjax -->
-  $$\text{max_speedup}_{\text{}} = \frac{1}{1 - \frac{query\_setup\_time}{query\_setup\_time+match\_time}}$$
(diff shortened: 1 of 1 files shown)


jobergum:
Sorry, but I don't follow you here. You write

>Amdahl's law is 1/(1-DQW/(DQW+SQW)) and not 1/(1-SQW/(DQW+SQW)) 

This is what is documented and where DQW/(DQW + SQW) gives you the p in Amdahl, the portion of the work that can be parallelized. A high portion of parallelizable work gives a higher speedup as (1 - p) approaches 0, and a low p approaches 1. 

It might be that it's too complex to bring in theorems,   It was written at a time when many apps tried to scale throughput by adding nodes to a flat layout, but there most of the cost was related to query setup, dictionary traversals so overall throughput did not improve at all.

Hadrien-Cornier:
There is no issue with the equation, the issue is that the latex code that you had does not reflect the equation accurately, so I changed it

jobergum:
Sorry, I now get it; there is another mentioning in the performance/sizing-search.html#metrics-for-vespa-sizing section which have the formula wrong :) Thanks!

PR #2495, opened 2023-01-24
Add reference doc for match profiling.

geirst:
This also includes parameters to turn on profiling for matching, first-phase ranking and second-phase ranking individually.

Support for match profiling is part of Vespa 8.114.40.

@havardpe and @kkraune please review
@jobergum @bratseth @baldersheim FYI


havardpe (approved):

kkraune (approved):
Great!

PR #2497, opened 2023-01-24
Pre-start script no longer supported

kkraune:
- this is removed from the most recent startup code

@bratseth this is an anti-feature, and can easily be implemented by users, so assuming good to remove


arnej27959 (approved):
👍

bratseth:
👍

PR #2502, opened 2023-01-26
* better examples for fieldsets in get/visit operations

arnej27959:
* update with correct (Vespa 8) default

@kkraune please review


kkraune (approved):
Great! Thanks!

kkraune:
you can see if my deprecation comment makes sense

PR #2503, opened 2023-01-26
Clarify rank profile evaluation

kkraune:


bratseth (changes requested):
It's correct that sum isn't evaluated lazily here, but this is still pretty much misleading. Rank profiles in themselves are not evaluated, only functions, and functions in rank profiles *are* evaluated lazily (and so are arguments to "if"). And I don't see where lambdas enters into this - assuming "bar" is invoked at all there's just nothing optional in it - and furthermore the "sum" in bar is a lambda.

bratseth (approved):

bratseth:
Disregard the above, I was correct after all 3 months ago, just hard to understand this example (imho).

bratseth:
Vespa is just like regular programming languages here - function arguments are evaluated as the function is invoked, not as the function argument value is accessed inside the function.

... so it may not be necessary to document it.

PR #2507, opened 2023-01-27
SEO query plan / hints

kkraune:
- from slack thread

### DIFF
--- a/en/query-api.html
+++ b/en/query-api.html
@@ -232,6 +232,10 @@ use this to write code to manipulate results.
   See <a href="performance/sizing-search.html#life-of-a-query-in-vespa">life of a query</a>
   for a deeper dive into query execution details.
 </p>
+<p>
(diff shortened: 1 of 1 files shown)


en/query-api.html
```diff
@@ -232,6 +232,10 @@ <h2 id="query-execution">Query execution</h2>
   See <a href="performance/sizing-search.html#life-of-a-query-in-vespa">life of a query</a>
   for a deeper dive into query execution details.
 </p>
+<p>
+  Use <a href="reference/query-api-reference.html#trace.explainlevel">trace.explainlevel</a> to analyze the query plan.
+  Vespa does not have a feature to add hints to modify the query plan.
```
jobergum, line 237: We sort of do, we have capped range search which impacts the query plan, also ranked:false query annotations.

kkraune: good points! let me add that. thanks!

kkraune:
@jobergum PTAL

PR #2523, opened 2023-02-13
refactor to use requests

jobergum:
Avoid using curl/popen which will perform an expensive TLS handshake for every request, plus the cost of a process fork. Last feed to all 3 regions took 2 hours and 27 minutes. 


kkraune (approved):
much better!

PR #2539, opened 2023-02-27
Add smoketest draft

kkraune:
- We need to change the triggering, this should be invoked as part of the build/release pipeline


aressem:
How should this be used to smoke test the image not yet published to `vespaengine/vespa` ?

PR #2545, opened 2023-03-01
primary ordering on the rank score

bjormel:

### DIFF
--- a/en/reference/query-language-reference.html
+++ b/en/reference/query-language-reference.html
@@ -1216,7 +1216,9 @@ where predicate(predicate_field,0,{"age":20L})
   Sort using <code>order by</code>.
   Add <code>asc</code> or <code>desc</code> after the name of an
   <a href="../attributes.html">attribute</a> to set sort order -
-  ascending order is default.
+  ascending order is default. Add another sorting attributes to get a secondary sort, that will be a tiebreaker for the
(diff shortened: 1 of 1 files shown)


en/reference/query-language-reference.html
```diff
@@ -1232,7 +1232,9 @@ <h2 id="order-by">order by</h2>
 <p>
   The <a href="../ranking.html">rank profile</a> determines the rank score each document will get.
   Results are ordered by that value by default, but <code>order by</code> overrides that ordering.
-  Vespa does not optimize away the rank score computation in this case, it is still executed, even if the model score is thrown away. Use the built-in rank-profile <em>unranked</em> for optimal performance of sorting queries.
+  Vespa does not optimize away the rank score computation in this case, it is still executed, even if the model score is thrown away. Use the built-in rank-profile <em>unranked</em> for optimal performance of sorting queries.<br>
```
kkraune, line 1235: Maybe clarify that the secondary is a tie-breaker if the relevance is equal?
```suggestion
  Vespa does not optimize away the rank score computation in this case, it is still executed, even if the model score is thrown away. Use the built-in rank-profile <em>unranked</em> for optimal performance of sorting queries.
  </p>
  <p>
```

kkraune (approved):

kkraune (approved):
👍

PR #2561, opened 2023-03-14
chown /opt/vespa, too

kkraune:
FYI @aressem

### DIFF
--- a/en/operations/docker-containers.html
+++ b/en/operations/docker-containers.html
@@ -20,12 +20,8 @@ redirect_from:
   show how to run Vespa in Docker containers.
   In these examples, all the data is stored inside the container - the data is lost if the container is deleted.
   When running Vespa inside Docker containers in production,
-  volume mappings to the parent host should be added to persist data and logs.
+  volume mappings to the parent host should be added to persist data and logs:
 </p>
-<ul>
-  <li>/opt/vespa/var</li>
-  <li>/opt/vespa/logs</li>
-</ul>
 <pre>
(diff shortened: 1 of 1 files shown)


aressem:
Rationale for this ? Seems wrong.

kkraune:
Testing on an EC2 instance with Docker: 

````
$ vespa-deploy prepare .
Uploading application '.' using http://localhost:19071/application/v2/tenant/default/session
Session 3 for tenant 'default' created.
mkdir /opt/vespa/.cloudconfig: permission denied
````

FYI @hmusum

aressem:
The vespa CLI also defaults to writing stuff in $HOME for the vespa user which is /opt/vespa, but for that tool we did set some env variables in the container. Maybe this is possible for vespa-deploy as well so we can avoid writing into /opt/vespa ?

kkraune:
I am all for that, not sure if other things write here, too, we will find out. what do you say, @hmusum ?

kkraune:
Maybe @arnej27959 have input too, he let me know we had to chown but I am not sure what was the original error - maybe user complaint?

arnej27959:
the original complaint was very confused but indicated some directories had wrong user.
we should document how to fix ownership of any volumes etc mounted into the docker container (if needed).
/opt/vespa should not need to be writable, so we should fix that failure somehow.

Various tools (including vespa-deploy) expect to be able to write to the home directory of the current user;
maybe we should add a /opt/vespa/home owned by vespa and have that as the home directory?

kkraune:
I have created https://github.com/vespa-engine/vespa/issues/26473 to fix the software instead

PR #2563, opened 2023-03-15
move details to reference, use vespa-cli visit

kkraune:
- this is half-way through, will do more once `vespa visit` supports more arguments - can clean up more then


vekterli (approved):
:+1:

PR #2572, opened 2023-03-21
Add unit reference doc. Rename metric set files

olaaun:


yngveaasheim (approved):
👍

kkraune (commented):
Great! Maybe we should add a comment in all the autogenerated files that they are in fact autogenerated from code, and where to look to update scripts etc

PR #2603, opened 2023-04-18
Kkraune/testing guide

kkraune:
- Improve the guide with examples moved from reference doc
- Do not mention prod testing from the guide, it is not really actionable here. Kept it in the reference


jonmv (approved):
:)

PR #2639, opened 2023-05-11
renamed innerproduct -> prenormalized-angular 

arnej27959:
@geirst please review
@kkraune FYI


kkraune (commented):
👍

geirst (approved):
👍 
There are still some mentioning of 'unit vectors' in the distance-metric reference doc for angular and prenormalized-angular. That can be lifted as well.

PR #2681, opened 2023-05-24
Document conditional Put with create-if-nonexistent

vekterli:
@kkraune please review. Should not be merged before the underlying feature is enabled by default (will probably patch in required Vespa version at that point).

Also clean up some use of explicit tables for layout and replace with equivalent Denali CSS classes.


kkraune (approved):
Great!

vekterli:
@kkraune the features required for conditional put+create to work are now enabled by default in the freshly squeezed Open-Source release. Added a few notes on supported version. PTAL.

kkraune:
Great - merging!

kkraune:
or you merge, the label is still there

PR #2707, opened 2023-06-09
Clarify the restrictions around document name and its equivalence to document type reference

rharige:

### DIFF
--- a/en/reference/schema-reference.html
+++ b/en/reference/schema-reference.html
@@ -258,7 +258,7 @@ document [name] inherits [name-list] {
 </pre>
 <p>
   The document name is optional, it defaults to the containing <code>schema</code>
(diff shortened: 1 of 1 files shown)


en/reference/schema-reference.html
```diff
@@ -258,7 +258,7 @@ <h2 id="document">document</h2>
 </pre>
 <p>
   The document name is optional, it defaults to the containing <code>schema</code>
- element's name. If there is no containing <code>schema</code> element, the document name is required.
+ element's name. If there is no containing <code>schema</code> element, the document name is required. If the document with a name is defined inside a schema, the document name must match the <code>schema</code> element's name. The reference to `document type` in the documentation refers to the document name defined here.
```
kkraune, line 261: better highlighting
```suggestion
 element's name. If there is no containing <code>schema</code> element, the document name is required. If the document with a name is defined inside a schema, the document name must match the <code>schema</code> element's name. The reference to <em>document type</em> in the documentation refers to the document name defined here. 
```

PR #2712, opened 2023-06-13
Simplify: single embedding per doc

kkraune:


jobergum (approved):

jobergum (approved):

jobergum:
Would it make sense to remove the green numbers? It's a pain to have both an illustration and a list that explains the illustration. I think the illustration is self-explanatory.

PR #2722, opened 2023-06-16
link to Parameter substitution

kkraune:


jobergum (approved):

jobergum:
Thanks for picking up this, I thought about it this week actually!

PR #2731, opened 2023-06-21
Remove alias from doc

kkraune:


hmusum:
I thought alias was a feature, are we dropping it?

hmusum (approved):
OK, found explanation in a ticket, looks good

hmusum:
Someone might want to know how to use another name for a field, I hope that is explained in doc

bratseth:
We shouldn't remove aliases (on fields), only creating indexes and attributes as logically separate entities with different names., They should always be named the same as the fields in which they are contained (and so of course also have the same aliases as it).

kkraune:
Good clarification, I will modify this PR

kkraune:
alias now only documented for field, with a simple example - @hmusum ptal

hmusum (approved):
👍

PR #2741, opened 2023-06-26
Document enabling indexing in streaming

bratseth:
And point to the streaming sample app.


kkraune (approved):
will try out the sample app for the newsletter :-) this looks great

geirst (commented):
👍

PR #2747, opened 2023-07-04
distanceThreshold is a double.

geirst:
@arnej27959 please review


en/reference/query-language-reference.html
```diff
@@ -1512,7 +1512,7 @@ <h2 id="annotations">Annotations</h2>
   <tr>
     <td>distanceThreshold</td>
     <td>2</td>
```
arnej27959, line 1514: "2" is wrong also

geirst: Fixed

arnej27959 (approved):
👍

PR #2751, opened 2023-07-06
Remove link to table row, it did not work properly.

gjoranv:


bjorncs (approved):

kkraune:
we did never find out why this never worked with the header/scrolling - when we link into rows, we use a &lt;p&gt; :

`<tr><th><p id="allowEmbeddedArtifacts">allowEmbeddedArtifacts</p></th>`

gjoranv:
Great tip, I'll remember it for later.  I think we'll keep it like this anyway, as I find it a bit clearer.

PR #2785, opened 2023-08-04
Add text on searcher timeout

kkraune:
- .. from slack conversation

### DIFF
--- a/en/searcher-development.html
+++ b/en/searcher-development.html
@@ -645,8 +645,7 @@ public class UnsafeSearcher extends Searcher {
 <h2 id="timeouts">Timeouts</h2>
 <p>
   How to gracefully handle a timeout inside a Searcher?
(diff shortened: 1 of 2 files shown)


en/searcher-development.html
```diff
@@ -642,6 +642,29 @@ <h2 id="error-handling">Error handling</h2>
 </p>
 
 
+<h2 id="timeouts">Timeouts</h2>
+<p>
+  How to gracefully handle a timeout inside a Searcher?
+  <code>Result result = execution.search(query)</code> can result in a timeout, and is not null.
```
jobergum, line 648: What does this mean - "and is not null" ?

jobergum (approved):

en/searcher-development.html
kkraune: search() returns a non-null object - but can remove this, not important.

PR #2789, opened 2023-08-07
- Remove ancient round-robin dispatch-policy from documentation. It h…

baldersheim:
…as too many caveats.

- Document best-of-random-2 dispatch-policy.

@kkraune PR
@geirst @yngveaasheim FYI


kkraune (approved):
👍

PR #2802, opened 2023-08-14
Add Vespa-CLI autogenerated ref doc MERGEOK

kkraune:
- Created by running `vespa gendoc .` + a few more steps for serving
- In next PR, add a screwdriver job that regularly runs `vespa gendoc` and checks for diff - and updates repo if so. Use update-vespa-version in https://github.com/vespa-engine/documentation/blob/master/screwdriver.yaml#L95 for inspiration.
- Finally, link to here from a few places

CC @jobergum

### DIFF
--- a/en/reference/vespa-cli/vespa.md
+++ b/en/reference/vespa-cli/vespa.md
@@ -1,0 +1,57 @@
+---
+render_with_liquid: false
+---
+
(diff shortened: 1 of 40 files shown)


mpolden (approved):
👍

PR #2805, opened 2023-08-15
Kkraune/cli auto doc

kkraune:
- trigger on pr and commit while developing, will trigger this weekly later, as the source changes are in vespa-engine/vespa
- some debug output for now, the escaping is hard to get right on first try ...

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -115,14 +115,64 @@ jobs:
             echo "Latest Vespa version is: " \${VESPA_VERSION}
             sed -i.orig "s/vespa_version: \"[0-9\.]*\"/vespa_version: \"\${VESPA_VERSION}\"/" _config.yml
             git diff
-            changes=\$(git status --porcelain | wc -l)            
+            changes=\$(git status --porcelain | wc -l)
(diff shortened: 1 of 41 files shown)


screwdriver.yaml
```diff
@@ -115,14 +115,64 @@ jobs:
...
+            sed -i.orig 's#\(https://docs.vespa.ai/[^\S]*\)#[\1](\1)#g' \*.md
+            rm \*.orig
+            git diff
+            changes=\$(git status --porcelain | wc -l)
+            if (( changes > 0 )); then
+              echo "Updating Vespa CLI reference documentation"
+              # git commit -m "Update Vespa-version to \${VESPA_VERSION} MERGEOK" _config.yml
+              # git pull --rebase
+              # git push
+            fi
```
mpolden, line 173: I think having the script in a separate file would be easier to work with.

mpolden (approved):

screwdriver.yaml
kkraune: agreed :-) there are more jobs like this, so could generalize

PR #2824, opened 2023-08-21
rewrite tensor user guide

jobergum:
Trying to be more concrete. 

- A complete schema example
- A complete feed json format of both mapped and indexed
- A complete query example for both vespa cli, get and post (since)


kkraune (approved):
💯  I like that it is _much_ more friendly to copy and paste! Great examples!

PR #2826, opened 2023-08-22
Kkraune/json ref

kkraune:
- The major changes are highlighting, removing items that are in /document/v1 (linking there instead) and reordering


jobergum:
Great

PR #2831, opened 2023-08-24
Use api.search.vespa.ai

kkraune:


jobergum:
Don't think that is open yet.

kkraune:
right, hold until then

PR #2833, opened 2023-08-25
Fix content typos

buinauskas:

While reading metrics documentation, I've noticed some typos - let's fix them.

### DIFF
--- a/en/approximate-nn-hnsw.md
+++ b/en/approximate-nn-hnsw.md
@@ -169,7 +169,7 @@ but this is less relevant as the `HNSW` index search first reduces the document 
 
 * **targetHits**:
 The [targetHits](reference/query-language-reference.html#targethits)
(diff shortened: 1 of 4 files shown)


kkraune:
Thanks! The metrics reference is actually generated from code, so the typo is in https://github.com/vespa-engine/vespa/blob/master/metrics/src/main/java/ai/vespa/metrics/ContainerMetrics.java#L130 - you can make a PR here. This first one can be merged as-is

buinauskas:
I did submit a change in https://github.com/vespa-engine/vespa/pull/28147, I assume I can close this one down then.

kkraune:
I will merge the full PR - as mentioned, the 3 files will be overwritten at next metrics ref generation, just make it comment here whether you will make the PR for that or I can do

kkraune:
ok, i see that now. merging this one too!

PR #2837, opened 2023-08-25
Kkraune/ranking query

kkraune:
- A little easier to use query inputs
- Link to content in tensor guide and query api reference


kkraune:
ping @jobergum

jobergum (approved):

PR #2838, opened 2023-08-25
Document "create": true for tensor modify update.

geirst:
@arnej27959 please review

Do not merge until Vespa 8.217.43 is released.


arnej27959 (approved):
looks good to me; maybe we should explicitly say that created cells end up with the value you send in as add/multiply factor.

geirst:
@arnej27959 PTAL

geirst:
Vespa 8.218.31 is now available.

PR #2854, opened 2023-09-04
Add minimal example, more links and clarifications

kkraune:
CC @geirst


jobergum (approved):

jobergum:
Great cleanup. We can even cut it down further with less prose and point more to the practical guide for usage.

PR #2863, opened 2023-09-06
Kkraune/text matching MERGEOK

kkraune:
- the gist of this PR is to have a good resource for text matching
- moved the text ranking items to the general ranking doc
- added more examples (regexp), re-organized to a more natural flow


jobergum (approved):
💯 See if you can update the synthetic questions paths when you re-name files.

kkraune:
Good point willdo

PR #2864, opened 2023-09-07
Inherits schema to inherit elements

bjormel:

### DIFF
--- a/en/schemas.html
+++ b/en/schemas.html
@@ -282,13 +282,27 @@ Applications can map many schemas to one content cluster.
 Use multiple content clusters if documents of different schemas have different performance characteristics -
 read more in the <a href="performance/sizing-search.html">serving scaling guide</a>.
 </p>
-<pre>
-&lt;content id="items" version="1.0"&gt;
-    &lt;documents&gt;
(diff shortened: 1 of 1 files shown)


en/schemas.html
```diff
@@ -270,11 +270,9 @@ <h2 id="multiple-schemas">Multiple schemas</h2>
 </p>
 <pre>
 $ENDPOINT/search/?
-  yql=select * from sources * where title contains "bob"&amp;
+  vespa query "select * from sources * where title contains \"bob\""&amp;
```
kkraune:
```suggestion
  vespa query "select * from sources * where title contains \"bob\""
```

en/schemas.html
```diff
@@ -195,7 +195,7 @@ <h3 id="fieldset">fieldset</h3>
 }
 
 $ENDPOINT/search/?
```
kkraune: remove the $ENDPOINT/search/? as well?

kkraune (approved):
Great! I have not tested schema inheritance myself, I assume this works well

bjormel:
hmm, since we have the this new AI powered doc search, I guess that we should use complete examples. Will add doc defs to schema examples where they are not defined

PR #2894, opened 2023-09-21
Use the notification colors

bratseth:
I don't think saturated bright green works well with this color scheme. We could find a green that works but just using the notification yellow is fine too ...


kkraune:
agreed, this is better for now

PR #2898, opened 2023-09-22
Add batch delete example

kkraune:


kkraune:
merging so end user can have a look

PR #2928, opened 2023-10-05
Document linguistics key

dainiusjocas:

In https://github.com/vespa-engine/vespa/pull/28682 the key for an analyzer was extended to support an optional `StemMode`. This PR documents the usage.

I've invented a term `linguistics key`. In the code the class is called `[AnalyzerKey](https://github.com/vespa-engine/vespa/blob/dd71247befb38345586504d49941665b1245257d/lucene-linguistics/src/main/java/com/yahoo/language/lucene/AnalyzerFactory.java#L164)`. If the `analyzer key` sounds like a better name, then I can change in docs.

Also, added 3 new supported languages introduced in https://github.com/vespa-engine/vespa/pull/28752.

Other minor readability improvements.


bratseth (approved):
👍

PR #2940, opened 2023-10-18
Update procedure for enabling bm25 with existing documents.

toregge:
@geirst : please review

### DIFF
--- a/en/reference/bm25.html
+++ b/en/reference/bm25.html
@@ -362,22 +362,17 @@ schema example {
 </p>
 
 <p>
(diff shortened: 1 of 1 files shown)


kkraune:
Great! So this means, you don't have to _do_ anything - just use the procedure to observe when proton is done?

geirst (approved):
👍

geirst:
> Great! So this means, you don't have to _do_ anything - just use the procedure to observe when proton is done?

Yes, this is handled automatically now. The observation via state/v1/custom/component was added for system testing and if for some reason monitoring is needed.

Available in Vespa 8.241.13.

PR #2947, opened 2023-10-20
Add field size guide entry

kkraune:
- @geirst the current reference doc says only the first 1M _characters_, please confirm this is correct (and not _terms_). This is not changed in this PR
- field size is a recurring question, so add a section in the Field section in Schema guide and link there

thanks @bratseth for slack input data :-)


geirst:
@kkraune the doc for `max-length` is correct (it limits characters).

geirst (approved):
👍

PR #2986, opened 2023-11-08
Stop recommending cmdline tools for debugging linguistic transformati…

geirst:
…ons.

This can now be done with the 'tokens' summary transform.

@kkraune please review
@toregge FYI


kkraune (approved):
Great! I think we tried to keep the commands alpha sorted (although they logically belong together)

PR #3002, opened 2023-11-21
Remove outdated notice on visitor semantics

vekterli:
@kkraune please review. Unless this refers to something I'm not aware of, this paragraph is very outdated. Additionally, the reprocessing example just above it still uses `vespa-visit` (not the Vespa CLI), so maybe it should be either rewritten or removed entirely?

Visiting has not sent document batch operations towards the client since the pre Vespa 5.1-days, so it should not be necessary to do anything special to avoid multioperation-related problems when reprocessing.


kkraune (approved):
great!

kkraune:
It is great if you have the time to rewrite to using `vespa visit` in a separate PR, too :-)

PR #3011, opened 2023-11-28
add bundle and close searcher tagg

andreer:


bjormel (approved):

bjormel:
@kraune fyi

gjoranv:
The intention is that this should not be necessary. The class in question should instead be added to `PlatformBundles:searchAndDocprocComponents`.

andreer:
Thanks @gjoranv, that makes this easier for users. If you can add it I can simplify the doc when it's in the latest open source release

PR #3019, opened 2023-12-05
Add a section about feeding back in data exported via visiting

vekterli:
@kkraune please review. I can't shake the feeling that this could be explained in more simple and/or understandable terms, but this is at least a start.
@jobergum FYI

### DIFF
--- a/en/visiting.html
+++ b/en/visiting.html
@@ -87,6 +87,30 @@ $ vespa visit --slices 100 --slice-id 0
 use <code>--bucket-space global</code> to visit these documents.
 By default, visiting only iterates over non-global documents.' %}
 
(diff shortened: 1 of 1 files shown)


en/visiting.html
```diff
@@ -87,6 +87,30 @@ <h2 id="data-export">Data export</h2>
 use <code>--bucket-space global</code> to visit these documents.
 By default, visiting only iterates over non-global documents.' %}
 
+<h3 id="feeding-data-exported-via-visiting">Performance note: feeding data exported via visiting</h3>
```
kkraune, line 90:
```suggestion
<h3 id="performance-note-feeding-data-exported-via-visiting">Performance note: feeding data exported via visiting</h3>
```

kkraune (approved):
👍

PR #3032, opened 2023-12-13
Bugfix query parameter name inputs => input

anagao:
alias is `input.query` per https://docs.vespa.ai/en/reference/query-api-reference.html#ranking.features


bratseth (approved):
Thanks!

PR #3062, opened 2024-01-18
Remove outdated note on conditional operation write-repair consistency

vekterli:
@kkraune please review

Conditional Put and Remove operations now feature proper handling of inconsistent replicas, with similar semantics to those of conditional Update operations.


kkraune (approved):
👍

PR #3069, opened 2024-01-22
Remove mention of broken "match" syntax for doc updates

jonmv:
@arnej27959 @vekterli please review.


jonmv:
Ok, let's keep it, then 😂

PR #3075, opened 2024-01-26
Document that character normalization is not done for streaming fuzzy/regex

vekterli:


kkraune (approved):
Great!!

PR #3078, opened 2024-01-29
Clarify character normalization semantics for regex/fuzzy

vekterli:
@geirst or @kkraune please review.

Fuzzy and regex matching not using character normalization applies to _both_ indexed and streaming search modes, so remove explicit mention of streaming search.


geirst (approved):
👍

kkraune (commented):
👍

PR #3083, opened 2024-02-02
Improve parameter substitution doc

kkraune:
- clean up reference doc, move examples to guide
- better example
- link to/from `embed`, which now support this

### DIFF
--- a/en/query-api.html
+++ b/en/query-api.html
@@ -167,6 +167,59 @@ $ vespa query "select * from music where userQuery()" \
 </p>
 
 
+<h3 id="parameter-substitution">Parameter substitution</h3>
+<p>
+  Parameter substitution lets you provide query values as request parameters
+  instead of inserting this into the YQL string itself.
+  This simplifies query generation, separating the value of the string/set/array from the YQL string -
+  i.e. the value will not corrupt the YQL string if it contains YQL-like syntax:
+</p>
+<ul>
+  <li>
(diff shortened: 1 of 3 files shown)


en/query-api.html
```diff
@@ -167,6 +167,59 @@ <h3 id="geo-filter-and-ranking">Geo Filter and Ranking</h3>
...
+... where !(id in (@my_set))&my_set=10,20,30
+
+# Use a weightedSet operator
+... where weightedSet(field, @my_set)&my_set={a:1,b:2}
+</pre>
+<p>Parameter substitution is also great to eliminate data duplication:</p>
+<pre>
+$ vespa query \
+  'yql=select id, from product where {targetHits:10}nearestNeighbor(embedding, query_embedding) or <span class="pre-hilite">userQuery()</span>' \
+  'input.query(query_embedding)=embed(transformer, userInput(<span class="pre-hilite">@query</span>))' \
```
jobergum, line 208: Is this correct usage?

kkraune: Corrected now, I think, by looking at https://github.com/vespa-engine/vespa/pull/29905

jobergum:
I'm worried about updating this without bold warnings on the Vespa version dependency because if people start using embed(@query) in a version that doesn't support it, it's hard to spot it, as you embed the string '@query' so that all query requests are embedded into the same point.

kkraune:
good point! Will add a note on version

jobergum (approved):

PR #3092, opened 2024-02-12
Document how to stream over multiple groups

bratseth:


en/reference/query-api-reference.html
```diff
@@ -1405,8 +1405,14 @@ <h2 id="streaming">Streaming</h2>
     <td></td>
     <td>
       <p id="streaming.selection">
-        Restricts streaming search using a selection instead of a group name.
-        This can be used to select a subset of documents based on any document selection expression.
+        Restricts streaming search using a selection expression instead of a group id.
+      </p>
+      <p>
+        If the selection is on the form <code>id.group == "foo" or id.group == "bar" or id.group == ...</code>
+        this will only stream documents in that group, which is efficient for a small number of groups.
```
vekterli, line 1412:
```suggestion
        this will only stream documents in those groups, which is efficient for a small number of groups.
```

vekterli (approved):
:+1:

PR #3094, opened 2024-02-14
Describe struct and struct-field with example

tmaregge:
Adds a brief description of struct and struct-field with examples with and without Pyvespa.


kkraune:
ping @geirst

PR #3109, opened 2024-03-10
Document using ColBERT embedding with chunks

bratseth:
Also general cleanup of	the embedding doc, which seems to have accumulated entropy faster than usual.

@jobergum or @kkraune

### DIFF
--- a/en/embedding.html
+++ b/en/embedding.html
@@ -6,68 +6,32 @@ redirect_from:
 ---
 
 <p>
-  A common technique in modern big data serving applications is to map unstructured data - say, text or images -
-  to points in an abstract vector space and then do computation in that vector space. For example, retrieve
+  A common technique is to map unstructured data - say, text or images -
+  to points in an abstract vector space and then do computation in that space.
+  For example, retrieve
   similar data by <a href="approximate-nn-hnsw.html">finding nearby points in the vector space</a>,
   or <a href="onnx.html">using the vectors as input to a neural net</a>.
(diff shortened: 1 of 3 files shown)


en/embedding.html
```diff
@@ -312,59 +380,125 @@ <h4 id="colbert-ranking">ColBert ranking</h4>
...
+
+<p>Transformer based models have named inputs and outputs that needs to be compatible
+with the input and output names used by the embedder.</p>
+
+<p>The <a href="https://github.com/vespa-engine/sample-apps/tree/master/simple-semantic-search#model-exporting">simple-semantic-search</a>
+sample app includes two scripts to export models and vocabulary files using the default expected input and output names
+for embedders using ONNX models. The input and output names to use can also be overriden by the various <code>transformer-</code>
+parameters in the <a href="reference/embedding-reference.html">config of the embedder in question</a>.
+
+<h3 id="onnx-debug">Debugging ONNX models</h3>
```
kkraune, line 395: overlaps with https://docs.vespa.ai/en/onnx.html#using-vespa-analyze-onnx-model - maybe the export/debug part here (that are great!) should be in the onnx doc instead, linked from here. I notice this PR does not add this, just rearrange, so can harmonize this in a follow-up PR

kkraune (approved):
💯

en/embedding.html
bratseth: Yes, agree that should be consolidated elsewhere.

jobergum (approved):
Wonderful 💯

PR #3130, opened 2024-03-28
Update searcher-development.html

PeterLombaers:
Update the example of writing a searcher to make use of the Vespa dataclasses.


bratseth (commented):
Thanks. I think this should be an example further down though - it's pretty far down on the list of things typically want to do and doesn't belong in the very first searcher you see.

PR #3141, opened 2024-04-10
Remove search input from content page

ldalves:
refs to VESPA-28445

### DIFF
--- a/index.html
+++ b/index.html
@@ -12,17 +12,6 @@ index: false
   </div>
 </div>
 
-<div class="row m-20">
-  <div class="col">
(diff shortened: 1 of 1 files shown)


kkraune (commented):
hm, we added this to make it more search centric, by @bratseth, but there might be a discussion I missed? I agree it is a little bit funny to have two, and I also think it is not using the correct suggestions, so we could improve. Not sure if we want to remove it

kkraune:
... and now I read the internal ticket, too - I assume you have discussed this, so approving for now

kkraune (approved):

PR #3145, opened 2024-04-17
Remove scorethreshold from WeakAnd

bratseth:
Not supported since Vespa 7


bratseth:
Pointed out in https://github.com/vespa-engine/vespa/issues/12179

geirst (approved):
👍

PR #3175, opened 2024-05-06
Add Diversity to Glossary

kkraune:


jobergum (approved):
Note that match-phase diversity is only relevant IF match-phase degradation is triggered.

PR #3235, opened 2024-06-20
Remove `distributor-base-port` from documentation

vekterli:
@kkraune please review.

This property likely hasn't worked for many years. My guess is that it's a leftover from the previous generation of content clusters.

At the very least we shouldn't document a broken feature.

This relates to https://github.com/vespa-engine/vespa/issues/31145


kkraune (approved):
Thanks!

PR #3264, opened 2024-07-23
Update Proton persistence layer documentation

vekterli:
@kkraune please review
@radu-gheorghe FYI

 * Remove outdated information on locking semantics.
 * Don't mention internal SPI concept that is not explained anywhere.
 * Add a fairly comprehensive overview of how operation scheduling, locking and throttling works at the content node persistence level.
 * Don't present having larger bucket split sizes as if it's something without inherent downsides as well.


kkraune (approved):
Great!

PR #3266, opened 2024-07-23
Link for zcurve in grouping reference

radu-gheorghe:
I'm hoping to clarify the meaning, especially since Wikipedia says "Not to be confused with [Z curve](https://en.wikipedia.org/wiki/Z_curve) or [Z-order](https://en.wikipedia.org/wiki/Z-order)."

Now I hope I'm linking the right one 😅


kkraune (approved):
👍

PR #3299, opened 2024-08-08
feat(link-checker): Simplify link-checker workflow configuration

esolitos:
## What

Simplify the link-checker workflow to utilize a reusable workflow. 

## Why

This reduces the amount of configuration needed in the repository and makes the workflow more maintainable.

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


esolitos:
😓 Test is failing because "https://cwiki.apache.org/" is timing out.. So it is correct that it should fail, but not related to my changes.

frodelu (approved):

PR #3318, opened 2024-08-19
Update details about which files and when files are distributed

hmusum:

### DIFF
--- a/en/application-packages.html
+++ b/en/application-packages.html
@@ -96,13 +96,15 @@ $ vespa prod deploy
 <h3 id="file-distribution">File distribution</h3>
 <p>
   The application package can have components and other large files.
-  At <em>vespa prepare</em>, these files are distributed to the nodes:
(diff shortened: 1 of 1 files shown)


en/application-packages.html
```diff
@@ -96,13 +96,15 @@ <h3 id="rollback">Rollback</h3>
 <h3 id="file-distribution">File distribution</h3>
 <p>
   The application package can have components and other large files.
-  At <em>vespa prepare</em>, these files are distributed to the nodes:
+  When an app is deployed (or more precisely, when the app is <em>activated</em>),
+  these files are distributed to the nodes:
```
kkraune, line 100: I learned something! I always thought prepare _distributed_ files to nodes, and activate _enabled_ them for use, this clarifies

kkraune (approved):

en/application-packages.html
bratseth: This is how it was initially, because activate should be fast and reliable, but at some point something happened ...

PR #3324, opened 2024-08-21
feat(services-container): Add link to environment variables example

esolitos:


kkraune:
still valid @esolitos ?

PR #3332, opened 2024-08-26
fix: add resource-limits under tuning

buinauskas:

The tuning section does not have a link to resource-limits, let's add it.


jobergum (approved):
Thanks a lot!

PR #3336, opened 2024-08-26
fix(ci): exclude www.ietf.org from the link-checker workflow

esolitos:
## What

- Excludes links to www.ietf.org to be checked

## Why

- It seems that they blocked access from CI systems, as the link [www.ietf.org/rfc/rfc2396.txt](https://www.ietf.org/rfc/rfc2396.txt) exists and is valid, but when checked via GH actions it returns 404.


jobergum (approved):
👍

PR #3348, opened 2024-08-29
Change some function signatures to make them more unified

theodorklauritzen:
This will make the documentation easier to integrate with the LSP

### DIFF
--- a/en/cross-encoders.md
+++ b/en/cross-encoders.md
@@ -152,7 +152,7 @@ rank-profile bert-ranker inherits default {
 }</pre>
 
 The example above limits the sequence length to `256` using the built-in 
-[convenience functions](reference/rank-features.html#tokenInputIds(length, input, ...)) 
+[convenience functions](reference/rank-features.html#tokenInputIds(length,input,...)) 
 for generating token sequence input to Transformer models. Note that `tokenInputIds` uses 101 as start of sequence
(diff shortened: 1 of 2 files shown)


kkraune:
while at it, I don't think space is a good idea in the id values, so wrong to begin with, so better replace the spaces with -

theodorklauritzen:
@kkraune I have removed the spaces in the ids, however I think the ids is easier to read without the dash.

kkraune:
Thanks! we have a convention to always lowercase, and replace whitespace/non-chars with dashes.

theodorklauritzen:
I've only changes four ids, and they are listed bellow. I think to insert dashes instead of spaces would be ugly in all of these cases. In the cases the dash should be right after a ','.

`tokenInputIds(length, input_1, input_2, ...)` -> `tokenInputIds(length,input,...)`

`customTokenInputIds(start_sequence_id, sep_sequence_id, length, input_1, input_2, ...)` -> `customTokenInputIds(startSequenceId,sepSequenceId,length,input,...)`

`tokenTypeIds(length, input_1, input_2, ...)` -> `tokenTypeIds(length,input,...)`

`tokenAttentionMask(length, input_1, input_2, ...)` -> `tokenAttentionMask(length,input,...)`

I have also changes the arguments from snake case to camel case, since everything else on this page is in camel case. Before the change all arguments with multiple words was in snake case. However there ware only four occurrences of snake case, and they are shown above.

PR #3349, opened 2024-08-29
fix(search): use the latest version of vespa-cli in the feed workflow

esolitos:
## What

use the latest version of vespa-cli in the feed workflow

## Why

There is no reason why we should use a specific version

---


esolitos:
Ignoring failures on Link checker as it still javadoc.io being down.

PR #3352, opened 2024-08-30
feat: Improve header parsing by removing backslashes

esolitos:
## What

- Follow up of #3335
- Removes escapes in the paragraph headers, this has been tested locally to generate the correct doc.

## Why

- This resolves a bug where dashes would be [escaped in the Search pages](https://search.vespa.ai/search?query=garbage-collection).

## Additional Info

Internal reference: [VESPANG-303](https://vespaai.atlassian.net/browse/VESPANG-303)

### DIFF
--- a/feed-split.py
+++ b/feed-split.py
@@ -100,7 +100,7 @@ def split_text(soup):
             if text:
                 data.append((id, header, text))
                 text = ""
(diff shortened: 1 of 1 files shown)


jobergum:
This is great @esolitos!

PR #3375, opened 2024-09-12
fix: services content min-node-ratio-per-group

ernestas-poskus:

There is a bug in `min-node-ratio-per-group` docs https://docs.vespa.ai/en/reference/services-content.html#min-node-ratio-per-group

> Default is 0, which means that the automatic group out-of-service functionality will <em>not</em> automatically take effect.

This is not correct with at least Vespa version 8.363.17. The default is 1 because the group was taken out of action when 1 node out of 20 was put in maintenance.

### DIFF
--- a/en/reference/services-content.html
+++ b/en/reference/services-content.html
@@ -1365,7 +1365,7 @@ Setting this tuning parameter allows the system to instead automatically take do
 allowing feed and query traffic to fail completely over to the remaining groups.
 </p><p>
 Valid parameter is a decimal value in the range [0, 1].
-Default is 0, which means that the automatic group out-of-service functionality will <em>not</em> automatically take effect.
+Default is 1, which means that the automatic group out-of-service functionality will automatically <em>take</em> effect.
 </p><p>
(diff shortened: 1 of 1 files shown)


ernestas-poskus:
Screenshot 2024-09-12 at 16 40 29

We have 3 content groups 20 nodes each, the search coverage settings are 
```xml
      <coverage>
        <minimum>0.8</minimum>
        <min-wait-after-coverage-factor>0.2</min-wait-after-coverage-factor>
        <max-wait-after-coverage-factor>0.3</max-wait-after-coverage-factor>
      </coverage>
```

We haven't had `min-node-ratio-per-group` set before. By observing the metrics panel I made the conclusion that default was actually 1 because if it was 0 then group would not be affected by the maintenance of 1 node.

> Containers will by default stop sending queries to a group (if there are better alternatives) if it has less than 97% coverage, and since 1/20 = 5% that applies here assuming each of your groups only have one active copy. See https://docs.vespa.ai/en/reference/services-content.html#min-active-docs-coverage if you want to tune this ratio.

This does makes sense but also setting the `min-node-ratio-per-group` to 0.8 should have not made any effect 🤔

PR #3390, opened 2024-09-20
Less controversial documentation example

johnphamvan:
Updating the Query Rewriting examples with a less controversial example that matches the original spirit. 

---


kkraune (approved):
Thanks!

PR #3417, opened 2024-10-10
doc-search.vespa.oath.cloud -> api.search.vespa.ai

freva:
AFAIK `js/search.js` and `js/search_suggestions.js` are not used anymore and we redirect to search.vespa.ai instead @kkraune ?


hakonhall (approved):

kkraune:
Good question, @freva ! we still link to the classic search experience, and https://docs.vespa.ai/search.html?q=ranking does not work when I try it now, not sure if related (just hangs on "Searching for 'ranking' ...")

freva:
Thanks @kkraune. There was an issue after the migration to api.search.vespa.ai, I've fixed it now.

PR #3427, opened 2024-10-16
Do not run doc testing for every push to a branch

kkraune:
Every little push to a branch is a 24-minute run, this is too excessive. The merge to master is enough for everyday use

### DIFF
--- a/.github/workflows/verify-guides-small.yml
+++ b/.github/workflows/verify-guides-small.yml
@@ -18,7 +18,7 @@ on:
     paths:
       - '.github/workflows/verify-guides.yml' # Reusable workflow
       - '.github/workflows/verify-guides-small.yml' # This workflow
-      
+
 jobs:
   small:
     uses: ./.github/workflows/verify-guides.yml
--- a/.github/workflows/verify-guides-small.yml
+++ b/.github/workflows/verify-guides-small.yml
@@ -13,6 +13,12 @@ on:
       - '.github/workflows/verify-guides.yml' # Reusable workflow
       - '.github/workflows/verify-guides-small.yml' # This workflow
 
+  pull_request:
+    branches: ["master"]
+    paths:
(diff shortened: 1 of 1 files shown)


esolitos:
🤔 I kinda disagree here, the test is there to ensure that if a guide is updated (or created) it will not be broken, the rightful place to be is in a PR.

At the _very very least_ we should leave it for changes to the workflows. 😊

kkraune:
I agree in _principle_ :-) In _practice_, very few PRs will change a test run. And we will see right after merge to master if it broke

Is there a middle ground here? Say in case, for a given PR, I want to trigger a run, because I am actually looking for a test run.

#3423 is an example of 90 minutes+ action run time for tiny doc changes ...

esolitos:
We could use the event [pull_request_review](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_review), so that this is triggered only when reviews are in.

Or else we could trigger it manually using a `/test` comment. 🤷🏼 

In either case the trigger on workflow changes should remain, because there we are directly affecting the workflows:
```yaml
  pull_request:
    branches: ["master"]
    paths:
      - '.github/workflows/verify-guides.yml' # Reusable workflow
      - '.github/workflows/verify-guides-small.yml' # This workflow
```

kkraune:
added back the workflow changes.

I did not know about the /test comment - need to learn more! maybe that is a practical tool whenever we change something that needs to be tested

esolitos:
Well the `/test` needs to be implemented in the workflow, does not "just work". :D

kkraune:
I am merging this for now. We can always add back the PR runs for testing. lets try without and see how it goes

kkraune:
BTW, I totally agree that we should have PR testing, but for the doc workflow we could look into something quicker, lets look at alternatives

PR #3459, opened 2024-10-31
explain more about Document ID

arnej27959:
@kkraune please review - some rewording may be required

### DIFF
--- a/en/documents.html
+++ b/en/documents.html
@@ -50,8 +50,8 @@ title: "Documents"
 
 <h3 id="id-scheme">id scheme</h3>
 <p>
-  Vespa has defined one scheme, the <em>id scheme</em>:
-  <code>id:&lt;namespace&gt;:&lt;document-type&gt;:&lt;key/value-pairs&gt;:&lt;user-specified&gt;</code>
+  Vespa currently has only one defined scheme, the <em>id scheme</em>:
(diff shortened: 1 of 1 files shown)


en/documents.html
```diff
@@ -99,8 +99,38 @@ <h3 id="id-scheme">id scheme</h3>
 </tbody>
 </table>
 
+<h3 id="docid-in-results">Document IDs in search results</h3>
+<p>
```
kkraune, line 103: link / refer to https://docs.vespa.ai/en/reference/default-result-format.html#id

kkraune (approved):

en/documents.html
```diff
@@ -99,8 +99,38 @@ <h3 id="id-scheme">id scheme</h3>
...
 
 <h3 id="namespace">Namespace</h3>
+<p>
+  The namespace is often just a placeholder.  Commonly used values are
+  a short form of the tenant or application name, or the string "ns".
+  One use case for different namespaces is when you have different
+  sources for documents.  For example, synthetic documents used for
+  testing can use namespace "test" so it's easy to detect and remove
+  them if they are present outside the test by mistake.
+</p>
```
bratseth, line 133:
```suggestion
<p>
  The namespace in document ids is useful when you have multiple document collections that you want 
  to be sure never end up with the same document id. It has no function in Vespa beyond this, and can
  just be set to any short constant value like for example "doc".
</p>
```

PR #3468, opened 2024-11-07
Document validation overrides

hmusum:

### DIFF
--- a/en/reference/validation-overrides.html
+++ b/en/reference/validation-overrides.html
@@ -48,9 +48,7 @@ title: validation-overrides.xml
 <h2 id="allow">allow</h2>
 <p>
   An <code>allow</code>-tag disables a particular validation for a limited time.
-  It contains a single validation-id,
-  see <a href="https://github.com/vespa-engine/vespa/blob/master/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationId.java">
-  ValidationId.java</a> for validation overrides.
+  It contains a single validation-id, see <a href="#list">list</a> below.
   <code>allow</code>-tags with unknown ids are ignored.
 </p>
 <table class="table">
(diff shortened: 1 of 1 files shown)


en/reference/validation-overrides.html
```diff
@@ -75,3 +73,101 @@ <h2 id="allow">allow</h2>
...
   </tbody>
 </table>
+
+
+<h2 id="list">List of validation overrides</h2>
+
+<table>
+  <tr>
+    <th>Validation id</th>
+    <th>Description</th>
```
esolitos, line 83:
```suggestion
    <th>Required when...</th>
```

en/reference/validation-overrides.html
```diff
@@ -75,3 +73,101 @@ <h2 id="allow">allow</h2>
 </tr>
   </tbody>
 </table>
+
+
+<h2 id="list">List of validation overrides</h2>
+
+<table>
+  <tr>
+    <th>Validation id</th>
```
esolitos, line 82:
```suggestion
    <th>ID</th>
```

I would have suggested "Validation Override ID" for clarity, but that's far to long for a single column.

en/reference/validation-overrides.html
```diff
@@ -75,3 +73,101 @@ <h2 id="allow">allow</h2>
...
+    <td>Changing zone (possibly private) endpoint settings.</td>
+    <td></td>
+  </tr>
+  <tr>
+    <td><code>redundancy-one</code></td>
+    <td>redundancy=1 requires a validation override on first deployment.</td>
+  </tr>
+  <tr>
+    <td><code>paged-setting-removal</code></td>
+    <td>May cause content nodes to run out of memory.</td>
```
esolitos, line 162: Could you add also a _"reason why anyone would set that"_ in there except the warning? 👀

en/reference/validation-overrides.html
```diff
@@ -75,3 +73,101 @@ <h2 id="allow">allow</h2>
...
+    <td></td>
+  </tr>
+  <tr>
+    <td><code>zone-endpoint-change</code></td>
+    <td>Changing zone (possibly private) endpoint settings.</td>
+    <td></td>
+  </tr>
+  <tr>
+    <td><code>redundancy-one</code></td>
+    <td>redundancy=1 requires a validation override on first deployment.</td>
```
esolitos, line 158:
```suggestion
    <td>The first deployment of an application with <code>redundancy=1</code> requires a validation override</td>
    <td></td>
```

Could we write (or link) an explanation "why" the first deployment with redundancy=1 requires a validation override?

en/reference/validation-overrides.html
```diff
@@ -75,3 +73,101 @@ <h2 id="allow">allow</h2>
...
+  </tr>
+  <tr>
+    <td><code>paged-setting-removal</code></td>
+    <td>May cause content nodes to run out of memory.</td>
+    <td>More data will be loaded into memory, might cause OOM.</td>
+  </tr>
+  <tr>
+    <td><code>certificate-removal</code></td>
+    <td>Remove data plane certificates.</td>
```
esolitos, line 167: Should we say "internal use" or explain why anyone would do such thing?

en/reference/validation-overrides.html
```diff
@@ -75,3 +73,101 @@ <h2 id="allow">allow</h2>
...
+    <td>Requires stopping all nodes, applying validation override and starting nodes again.</td>
+  </tr>
+  <tr>
+    <td><code>config-model-version-mismatch</code></td>
+    <td>For internal use.</td>
+    <td></td>
+  </tr>
+  <tr>
+    <td><code>skip-old-config-models</code></td>
+    <td>For internal use.</td>
```
esolitos, line 138: We should make a separate list after with only "reserved for internal use" to keep them separate, what do you think?.

esolitos (changes requested):
Nice! Overall looks great, however I'd suggest to tweak the descriptions a little but while we're at it so that is clear why would someone use that flags. In most places this is clear, but I highlighted a few spots where we should clarify.

en/reference/validation-overrides.html
hmusum: Yes, good idea, will do

en/reference/validation-overrides.html
hmusum: OK, will add some more

hmusum:
@esolitos please take a look again

esolitos (approved):

en/reference/validation-overrides.html
```diff
@@ -75,3 +73,113 @@ <h2 id="allow">allow</h2>
...
+  </tr>
+  <tr>
+    <td><code>certificate-removal</code></td>
+    <td>Removing data plane certificates, typically when moving to new certificates.</td>
+    <td>Unable to accesss endpoint with removed certificates.</td>
+  </tr>
+</table>
+
+
+<h3 id="list-internal">Validation overrides for internal use</h3>
```
kkraune, line 159: For end users, "internal use" is confusing. It might be easier to not expose this. To me, it is OK to keep the link to the source code, so easier to get to this

hmusum: Yeah, probably right. I'll use a link to source code instead. PR coming up

PR #3475, opened 2024-11-11
Ruby 3.2

kkraune:
- required by `activesupport-8.0.0 requires ruby version >= 3.2.0, which is incompatible with`


kkraune:
merging to see how it goes

PR #3477, opened 2024-11-12
first cut binarizing guide

kkraune:
- probably needs a few iterations, but can start off with this


jobergum (approved):
This is a great addition and very educational!

PR #3498, opened 2024-11-22
fix: Add parentheses around a annotation to fix a syntax error

theodorklauritzen:

The current example doesn't parse and run. Vespa gives the following error: `Could not create query from YQL: query:L1:114 extraneous input '\"c\"' expecting {<EOF>, 'select', ';'}`


kkraune (approved):
Thanks!

PR #3508, opened 2024-12-02
Use same rate.idDimension for the three different clients

frodelu:
@bratseth : Please review and merge. My understanding is that rate.idDimension is the dimension name used for the metrics emitted. Makes little sense to have this be different for each client?


bratseth:
Yes.

PR #3516, opened 2024-12-05
Mention that tensor fields can only be compared against null in document selections

vekterli:
@kkraune please review


kkraune (approved):
👍

PR #3545, opened 2024-12-19
add custom grouping queries as url parameters so examples can be shared by the url

lperry25:
- add custom grouping queries as url parameters so examples can be shared by the url
- add right margin to all pages to match the left margin
- trigger query when you hit enter inside input
- 
https://github.com/user-attachments/assets/5f21ce2e-3503-417c-844f-efe6bce368d4

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -39,7 +39,7 @@
       </div>
       {% endunless %}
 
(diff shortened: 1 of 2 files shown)


kkraune (approved):
nice - lets try this!

PR #3546, opened 2024-12-19
POC for a Vespa version select in the documentation

lperry25:
Create a POC for a Vespa version select in the documentation
- Selecting the version hides unrelated documentation (only one section has been added to selfhosted so far)
- This POC only considers entire pages that we could hide or show in the side bar. If the user navigates to the exact page they will still see it
- Logic could be added to hide or swap specific sections of the page

To allow merging and testing of this, the select will only appear if you add `?mode=cloud` or `?mode=selfhosted`. To go back to the docs without the select clear the url params from the URL and clear local storage

Screenshot 2024-12-19 at 16 18 42

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ GEM
   remote: https://rubygems.org/
   specs:
     Ascii85 (1.1.1)
-    activesupport (8.0.0)
+    activesupport (7.2.2.1)
       base64
       benchmark (>= 0.3)
       bigdecimal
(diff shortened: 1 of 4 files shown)


kkraune:
very cool! maybe we should just add this as new pages, so we can have index-new use navbar-new / sidebar-new so we can skunkwork this on the side without messing up the real site. Easy to experiment then!

kkraune (approved):
lets try this :-)

PR #3554, opened 2025-01-02
Kkraune/doc selection

kkraune:
- only the two last commits are relevant to review
- I _seem_ to remember that in multi-schema apps, visiting is faster is selecting on schema?


vekterli (commented):
:+1:

PR #3562, opened 2025-01-05
fix: wrong parameter name: not `container` but `jdisc`

dainiusjocas:

The param seems to be called `jdisc` and not `container`. The source of the config class is [here](https://github.com/vespa-engine/vespa/blob/f6acdd12bb24f65c4b3a036dea79cddfd1f13dd4/container-search/src/main/resources/configdefinitions/search.config.qr-start.def#L44).


bratseth (approved):
👍

PR #3579, opened 2025-01-14
Correct default feeding.concurrency

yjtakamabe:
correct default feeding.concurrency.

#### proton.def
```
feeding.concurrency double default = 0.2 restart
```
https://github.com/vespa-engine/vespa/blob/v8.461.16/configdefinitions/src/vespa/proton.def#L473

### DIFF
--- a/en/reference/services-content.html
+++ b/en/reference/services-content.html
@@ -905,7 +905,7 @@ Tune <a href="../proton.html">proton</a> settings for feed operations. Optional 
 </p>
 <ul>
   <li id="feeding-concurrency"><code>concurrency</code>:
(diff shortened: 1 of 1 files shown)


geirst (commented):
The default feed concurrency is actually 0.5 as seen here: https://github.com/vespa-engine/vespa/blob/master/flags/src/main/java/com/yahoo/vespa/flags/Flags.java#L109. 

This is used to set the proton config here: https://github.com/vespa-engine/vespa/blob/master/config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java#L163

yjtakamabe:
Thank you for reviewing

PR #3607, opened 2025-01-30
Add note on Jekyll force-polling to ensure docs are rebuilt on local changes

vekterli:
@kkraune please review
@geirst FYI

Although this _shouldn't_ be required, it seems this currently needs to be specified for automatic rebuilds to work, at least when running under macOS+Podman with a mapped directory.


kkraune (approved):
👍

PR #3614, opened 2025-02-04
more -> fewer

bratseth:


havardpe (commented):
'more' is correct

bratseth:
Don't we drop unconditionally based on term frequency here? If so this can really go either way right?

havardpe:
Yes. I suggest we add the word 'might' as well. The important point is that you adjust this limit to improve performance, but you gain a bit less than you would expect because you generally get more results. This is when you also need to look at the adjust limit that will help you reduce the number of results produced.

There are some inherent assumptions that (#docs * stopword-limit) is much larger than target-hits and that (#docs * adjust-limit) is also somewhat larger than target-hits. The conclusion being that in prod when Vespa and weakand makes sense you typically get more results, but if you specifically test the feature with a small corpus you get fewer results. And as you say, specific queries can give fewer results, for example searching for a word matching a single document together with a stop word.

bratseth:
Thanks. https://github.com/vespa-engine/documentation/pull/3618

PR #3615, opened 2025-02-04
Add glossary entries for (normalized) document frequency and estimated hit ratio

vekterli:
@geirst please review


geirst (approved):

kkraune (commented):
This is a good place to put explanations like this, and super useful for users!

PR #3625, opened 2025-02-10
Doc how to expose schema info

bratseth:
I've answered this enough times ...


kkraune:
looking at it now, will add a fix or two

kkraune:
Great - that was not much code! we could maybe promote custom components like this more ...

kkraune (commented):
👍

kkraune (approved):

PR #3627, opened 2025-02-10
apply autocorrect + some format

kkraune:
- This PR is best reviewed per commit  - disregard the first, this is the autocorrect for better language, and non-functional

### DIFF
--- a/en/performance/feature-tuning.html
+++ b/en/performance/feature-tuning.html
@@ -100,10 +100,11 @@ for optimal performance. See reference <a href="../reference/schema-reference.ht
   typically using the <a href="../using-wand-with-vespa.html#weakand">weakAnd</a> query operator with <a href="../reference/bm25.html">BM25</a> ranking.
 </p>
 <p>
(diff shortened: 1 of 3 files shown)


en/performance/feature-tuning.html
```diff
@@ -183,7 +187,7 @@ <h2 id="tuning-query-performance-for-lexical-search">Tuning query performance fo
       <strong>Avoid using large posting lists all together</strong>:
       Setting <a href="../reference/schema-reference.html#weakand-stopword-limit">stopword-limit</a> to 0.6,
       ensures that all terms that are estimated to occur in more than 60% of the documents are considered stopwords
-      and dropped entirely from the query, also from ranking.
+      and dropped entirely from the query and also from ranking.
```
kkraune, line 190: @geirst : Some might ask what happens with stopwords in phrase search here, if parts of the phrase is dropped. but this is for a weakAnd, so that does not apply. we can consider clarifying this somewhere in the text, someone will ask about this sometime ...

kkraune:
@geirst please take a look at last commit. this is great doc, I just added some links and we finally have some docs on a posting list! we might link to this from other places in doc as well.

PR #3633, opened 2025-02-12
Kkraune/grouping maps

kkraune:
- the ref doc for maps / array of struct was unclear, cleaned up
- added and example for grouping over a map field
- FYI @bjorncs  and @radu-gheorghe 

this is part of a larger sweep. I will redo some of the `matched-elements-only` pieces, too, for now I stuffed an example in there so we have it. This will be separated to a better example and relate to the "how to reduce data size in results" by @andreer 

Related commits:
- https://github.com/vespa-cloud/vespa-documentation-search/pull/183
- https://github.com/vespa-engine/sample-apps/pull/1649 and https://github.com/vespa-engine/sample-apps/pull/1652

### DIFF
--- a/en/grouping.html
+++ b/en/grouping.html
@@ -2028,7 +2028,51 @@ document-summary name_only {
 
 
 
-<h2 id="grouping-over-map-field">Grouping over Map field</h2>
+<h2 id="grouping-over-a-map-field">Grouping over a Map field</h2>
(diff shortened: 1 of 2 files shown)


kkraune:
I merge this for now, for the next pass, but the reference should be more readable now, will get cranking on the examples

PR #3635, opened 2025-02-13
Add

bratseth:


kkraune (approved):
👍

kkraune:
We could consider moving this into the performance section. nice workaround for the memory restrictions :-)

PR #3651, opened 2025-02-18
Update Slack invite link

radu-gheorghe:
Should be the same one as in https://vespa.ai/developer/


kkraune:
maybe better to just use https://slack.vespa.ai/ ?

radu-gheorghe:
Yes, thanks! Though I couldn't reuse this branch&PR, so there's https://github.com/vespa-engine/documentation/pull/3652 for it.

PR #3659, opened 2025-02-24
Add note about using streaming and indexed mode in same cluster

hmusum:


vekterli (approved):
:+1:

en/streaming-search.html
```diff
@@ -28,6 +28,11 @@
 subset (say, users with huge amounts of data) by automatically sharding such data groups
 over multiple content nodes, searched in parallel.</p>
 
+<p>Note: Using both streaming and indexing mode in the same cluster is discouraged.
```
vekterli, line 31: nit: "indexing" should probably be "indexed" here to match what's in `services.xml`

hmusum: Yes, thanks, the 2 mode was what intended, let me fix

vekterli (approved):

PR #3673, opened 2025-02-28
Remove reference to tfrank.py, as it does not exist

chrisbolin:
The tutorial [text-search-ml](https://docs.vespa.ai/en/tutorials/text-search-ml.html) mentions a python script `tfrank.py`:

> To illustrate this, we trained linear models using the [TF-Ranking framework](https://github.com/tensorflow/ranking). The framework is built on top of TensorFlow and allow us to specify pointwise, pairwise and listwise loss functions, among other things. The following script was used to generate the results below (just remember to increase the number of training steps when using the script).

> `./src/python/tfrank.py`

However, that file is not found anywhere in [text-search/python](https://github.com/vespa-engine/sample-apps/tree/master/text-search/python) or even in the parent repo [vespa-engine/sample-apps](https://github.com/vespa-engine/sample-apps/).

### DIFF
--- a/en/tutorials/text-search-ml.md
+++ b/en/tutorials/text-search-ml.md
@@ -437,15 +437,6 @@ which takes the entire ranked list into consideration when updating the model pa
 To illustrate this, we trained linear models using the [TF-Ranking framework](https://github.com/tensorflow/ranking).
 The framework is built on top of TensorFlow and allow us to specify pointwise, pairwise and listwise loss functions,
 among other things.
-The following script was used to generate the results below
-(just remember to increase the number of training steps when using the script).
-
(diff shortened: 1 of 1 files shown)


kkraune:
woops, sorry about that - we normally have testing steps in our guides to catch such issues, that is missing here - your PR makes sense then, thanks for submitting!

chrisbolin:
@kkraune if there is a version of `tfrank.py` around that could be integrated into the docs, let me know. Happy to take a look

kkraune:
Thanks! unfortunately, could not find any traces of it, this is years back. We should take another pass at this sample app

PR #3674, opened 2025-03-01
To configure OpenNlpLinguistics, it must be declared as a component

arnej27959:
The documented way to configure createCjkGrams had no effect. Looking at the systemtest it seems it's necessary do declare a component in services.xml - not sure if this is supposed to be necessary.
This PR follows the recommended way to configure components from https://docs.vespa.ai/en/configuring-components.html and I've tested that it actually takes effect.

@bratseth please review
@kkraune @bjormel FYI

### DIFF
--- a/en/linguistics.html
+++ b/en/linguistics.html
@@ -91,9 +91,11 @@ width="810px" height="auto" />
 <pre>
   &lt;container id="..." version="1.0"&gt;
     ...
-    &lt;config name="ai.vespa.opennlp.open-nlp"&gt;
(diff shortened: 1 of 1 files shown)


bratseth:
Good catch ... I'll update the system test.
But I think we should fix it instead: https://github.com/vespa-engine/vespa/pull/33459

kkraune:
Is this still useful with the Vespa change / should it be merged or closed?

bratseth:
Should be closed I think.

PR #3734, opened 2025-03-26
improve the lineheight on cloud only tooltip and remove some of the extra margin

lperry25:
Improved version looks like:

Screenshot 2025-03-26 at 09 38 36


kkraune (approved):
👍

PR #3772, opened 2025-04-09
Document how to go back to using original field after reindexing

hmusum:


kkraune (approved):
Great! this matches what Arne told me, too, makes sense!

PR #3787, opened 2025-04-15
Add toc-generation

vadimkatsel:
## Main changes
- add new `toc.html` to the includes
- add a call of a new TOC-generator `in default.html`
- add styles to `style.scss` for `toc`
- refactor `faq.md`(since it was having its own TOC which wasn't compatible with a new one)
## **Little preview**
image

### DIFF
--- a/_includes/htmlhead.html
+++ b/_includes/htmlhead.html
@@ -30,7 +30,7 @@
     <link rel="stylesheet" href="{{ site.baseurl }}/css/style-light.css">
     {% endif %}
     <!-- Table converter -->
-    {% if page.content contains '<!-- depends on mathjax -->'%}
+    {% if page.content contains '<!--table-to-list-->'%}
     <script src="https://code.jquery.com/jquery-3.7.1.min.js"></script>
     <script src="/js/table-list-converter.js"></script>
     {% endif %}
--- a/_includes/htmlhead.html
+++ b/_includes/htmlhead.html
@@ -29,6 +29,11 @@
     {% if include.theme == "light" %}
     <link rel="stylesheet" href="{{ site.baseurl }}/css/style-light.css">
     {% endif %}
+    <!-- Table converter -->
+    {% if page.content contains '<!-- depends on mathjax -->'%}
+    <script src="https://code.jquery.com/jquery-3.7.1.min.js"></script>
+    <script src="/js/table-list-converter.js"></script>
+    {% endif %}
 
     <!-- MathJax for formulas -->
     <script src="https://cdnjs.cloudflare.com/polyfill/v3/polyfill.min.js?features=es6"></script>
--- a/_includes/toc.html
+++ b/_includes/toc.html
@@ -1,29 +1,13 @@
 <!-- Copyright Vespa.ai. All rights reserved. -->
 {% assign h2_count = content | split: '<h' | size | minus: 1 %}
     {% if h2_count > 4 %}
+  
 
-    {% if page.title == 'FAQ - frequently asked questions'%} 
-    <style>
-    /* Fixed widths per heading level */
-    .toc ul .toc-level-0 > a {
-        width: 100% !important;
-    }
-    
-    .toc ul .toc-level-1 > a {
-      width: 90% !important;
-    }
-    
-    .toc ul .toc-level-2 > a {
-      width: 80% !important;
-    }
-    </style>
-    {% endif %}
-
-<nav class="toc" aria-labelledby="on-this-page-heading" style="position: fixed; align-self: flex-start;">
-    <h3 class="row" style="pointer-events: none;">On this page:</h3>
+<nav class="toc {% if page.title == 'FAQ - frequently asked questions'%} faq-toc {% endif %}" aria-labelledby="on-this-page-heading" style="position: fixed; align-self: flex-start;">
+    <h3 id="on-this-page-heading"  style="pointer-events: none;">On this page:</h3>
     {% assign headings = content | split: '<' %}
     
-    <ul style=" z-index: 3; max-height: 80vh; overflow-y: auto;">
+    <ul>
           {% for tag in headings %}
           {% assign tag_slice = tag | slice: 0, 1%}
           {% if tag_slice == '/' %}
--- a/_includes/toc.html
+++ b/_includes/toc.html
@@ -1,0 +1,45 @@
+<!-- Copyright Vespa.ai. All rights reserved. -->
+<div class="toc" style="position: fixed;">
+
+    {% assign headings = content | split: '<' %}
+      {% if headings %}
+      <h2 class="row">On this page:</h2>
+      {% endif %}
+    <ul style=" z-index: 3; max-height: 80vh; overflow-y: auto;">
+          {% for tag in headings %}
+          {% assign tag_slice = tag | slice: 0, 1%}
+          {% if tag_slice == '/' %}
+            {% continue %}
+          {% else %}
+            {% if tag contains 'h1' or tag contains 'h2' or tag contains 'h3' %}
+              {% assign toc_level = tag | slice: 1, 2%}
+              {% assign parts = tag | split: '>' %}
+              {% assign heading_id_first_cut = parts[0] | split: 'id="' %}
+              {% assign heading_id_second_cut = heading_id_first_cut[1] | split: '"' %}
+              {% assign heading_id = heading_id_second_cut[0] | strip | lstrip %}
+              {% assign heading_text = parts[1] | strip | lstrip %}
+  
+              {% if heading_text != "" %}
+              <li class="toc-level-{{toc_level | minus: 1 }} m-l-20 m-r-20">
+                <a href="#{{ heading_id }}">{{ heading_text }}</a>
+              </li>
+          {% endif %}
+            {% endif %}
+          {% endif %}
+          {% endfor %}
+  
+    </ul>
+  </div>
+  
+  
+  <script>
+      document.addEventListener("DOMContentLoaded", function() {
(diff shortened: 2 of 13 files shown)


kkraune:
this is a great start! a few comments

* we should have a variable in the frontmatter to disable (e.g. the docs frontpage should be without this)
* I'd like a more neutral style, just the text, no icons or boxes
* there is much whitespace to the right, and at the same time, the main text is narrow.
* it should disappear before the left menu when I shrink the window width
* do not show heading if empty, ref /en/tensorflow.html
* smaller "on this page"
* /en/reference/rank-features.html looks wrong, but this page is not right today, either, it is too wide. this doc has a giant table, maybe time to find a better format for this ...

vadimkatsel:
Thanks for the review! I'll change it.

vadimkatsel:
This is the list of what were done:
- now toc appears only if h2(sub-titels) provided.
- more neutral style were applied: no icons or boxes.
- whitespace were caused by settin "max-width" for the central row. Grid fix were provided.
- Now It disappears one level before the left menu
- smaller "on this page"
- /en/reference/rank-features.html is still looking wrong, haven't found nice solution yet.

_includes/toc.html
```diff
@@ -0,0 +1,44 @@
+<!-- Copyright Vespa.ai. All rights reserved. -->
+
+
+<div class="toc" style="position: fixed;">
+    <h3 class="row">On this page:</h3>
```
lperry25, line 5: We should disable the hover on this title, since I don't think it makes any sense to copy a link to this nav section
Screenshot 2025-04-23 at 09 37 19

_includes/toc.html
```diff
@@ -0,0 +1,44 @@
+<!-- Copyright Vespa.ai. All rights reserved. -->
+
+
+<div class="toc" style="position: fixed;">
```
lperry25, line 4: For best accessiblity we should wrap this section in a nav tag and an aria-labelled-by pointing to the header of the section. I notice we don't have a nav section on our main site nav either which will also help accessibility. For a screen reader, it becomes quite difficult to navigate between our 3 nav bars now if we don't label them correctly.

```
<nav aria-labelledby="on-this-page-heading">
      <h3 class="row" id="on-this-page-heading"> 
....

```

_includes/toc.html
```diff
@@ -0,0 +1,44 @@
...
+    <h3 class="row">On this page:</h3>
+    {% assign headings = content | split: '<' %}
+    
+    <ul style=" z-index: 3; max-height: 80vh; overflow-y: auto;">
+          {% for tag in headings %}
+          {% assign tag_slice = tag | slice: 0, 1%}
+          {% if tag_slice == '/' %}
+            {% continue %}
+          {% else %}
+            {% if tag contains 'h1' or tag contains 'h2' or tag contains 'h3' %}
```
lperry25, line 14: Maybe we should only include the on this page section if there are a significant amount of sections. I think this page looks quite strange with only one heading. Maybe there should be a minimum of 3? Otherwise I think it adds little benefit for the extra 'mess' on the page
Screenshot 2025-04-23 at 09 36 54

_includes/toc.html
```diff
@@ -0,0 +1,44 @@
...
+              </li>
+          {% endif %}
+            {% endif %}
+          {% endif %}
+          {% endfor %}
+  
+    </ul>
+  </div>
+  
+<script>
```
lperry25, line 34: Is this script necessary? It seems like we should already have heading ids on each section since the sections already include links right?

_layouts/default.html
```diff
@@ -52,11 +47,17 @@ <h1>{{ page.title | escape }}</h1>
...
       {% endunless %}
 
-      <div class="m-l-20 m-r-20" style="max-width: 1000px">
+      <div class="m-l-20 m-r-20">
         {{ content }}
       </div>
-
     </div>
+    {% if content contains 'h2' %}
+    <div class="xs-hide sm-hide md-hide lg-col-3-12 xl-col-2-12">
```
lperry25, line 55: The section text gets cut off on some screen sizes right now. You can't scroll when this happens, so it's impossible to read the section names
Screenshot 2025-04-23 at 09 35 32

lperry25 (changes requested):

_includes/toc.html
vadimkatsel: Completly unnecessary. Nice observation! Thanks!
 To be removed with the next commit.

_includes/toc.html
vadimkatsel: Will be included in the next commit.

_includes/toc.html
vadimkatsel: Thank you! Figured out that it's possible to split on '<h' and count the number of headings in the content section, which means to be fixed with next commit.

_layouts/default.html
vadimkatsel: This one is tricky need some time to think a bit. It seems It's collapsing, when the content section is somehow bigger than expected by grid. in other word, grid need a closer look i suppose. Would be interesting to hear your opinion on what is the possible way of thinking here.

_layouts/default.html
```diff
@@ -52,11 +47,17 @@ <h1>{{ page.title | escape }}</h1>
       </div>
       {% endunless %}
 
-      <div class="m-l-20 m-r-20" style="max-width: 1000px">
+      <div class="m-l-20 m-r-20">
```
lperry25, line 50: To avoid long words breaking the content like we found here add these styles to this div
```
{
  word-break: break-word; 
  overflow-wrap: break-word;  
  word-wrap: break-word; 
}```

File: _layouts/default.html [hunk 7]
```diff
@@ -52,11 +47,17 @@ <h1>{{ page.title | escape }}</h1>
       </div>
       {% endunless %}
 
-      <div class="m-l-20 m-r-20" style="max-width: 1000px">
```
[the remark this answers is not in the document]
lperry25: Also you can add a style of `overflow-x: auto` so that we never have this issue with the section getting too large and breaking the view

kkraune:
work note: the FAQ work look OK - but I think we should do the following:
* show the full question in the TOC at right (now you must hover over it). So we need a long-form TOC type for this
* we can break the FAQ into more documents - this is easy with the new left-frame index - so we will have topic-specific FAQs. With this, the long-form TOC will be OK as there will be much less entries

The second can be in a PR separate from this, this PR is getting long enough as it is

vadimkatsel:
Let's hope It's final commit in this PR. It includes:

- Toc fixes (it was parsing some pages wrong: some scripts, which worked with h2 and h3 tags were included to parsed tags or some content containing 'h2' as `h2load benchmarking tool`, for example.
- width fixes: now the indicated pages are adaptable for any screen size.
- formula display fixes: now formulas scales with the screen size.


Thanks @lperry25 a lot! She saved a lot of our nerves helping here!
And thanks to @kkraune for testing it, allowing to make it as good as it possible.

_includes/htmlhead.html
```diff
@@ -29,6 +29,11 @@
     {% if include.theme == "light" %}
     <link rel="stylesheet" href="{{ site.baseurl }}/css/style-light.css">
     {% endif %}
+    <!-- Table converter -->
+    {% if page.content contains '<!-- depends on mathjax -->'%}
```
lperry25, line 33: is there any reason to not just include this for all the pages? I don't love the idea of basing the logic on a seemingly unrelated comment in the code

_includes/toc.html
```diff
@@ -0,0 +1,54 @@
+<!-- Copyright Vespa.ai. All rights reserved. -->
+{% assign h2_count = content | split: '<h' | size | minus: 1 %}
+    {% if h2_count > 4 %}
+
+    {% if page.title == 'FAQ - frequently asked questions'%} 
+    <style>
+    /* Fixed widths per heading level */
+    .toc ul .toc-level-0 > a {
+        width: 100% !important;
```
lperry25, line 9: is the !important required in these three lines? It's best practice to avoid important since it can be quite frustrating if trying to add specific style to something. Usually a css selector that is this specific should be enough without the use of important

_includes/toc.html
```diff
@@ -0,0 +1,54 @@
...
+    }
+    
+    .toc ul .toc-level-2 > a {
+      width: 80% !important;
+    }
+    </style>
+    {% endif %}
+
+<nav class="toc" aria-labelledby="on-this-page-heading" style="position: fixed; align-self: flex-start;">
+    <h3 class="row" style="pointer-events: none;">On this page:</h3>
```
lperry25, line 23: you're missing the id="on-this-page-heading" for the aria-labelledby you defined in the line above

lperry25: why not add `position: fixed; align-self: flex-start;` to the toc class instead of inline styles? class styling is typically preferred

_includes/toc.html
```diff
@@ -0,0 +1,54 @@
...
+    {% assign headings = content | split: '<' %}
+    
+    <ul style=" z-index: 3; max-height: 80vh; overflow-y: auto;">
```
lperry25, line 26: can these inline styles be added the `.toc ul` styles you have defined in style.css? or does this ul have a different purpose than general ul inside toc? If so then I think it is more clear when reading the code if you define these as a specific class name rather than inline styling

css/style.scss
```diff
@@ -747,3 +747,123 @@ blockquote {
...
+  }
+
+
+  /* Container overflow */
+
+  .container-overflow {
+    overflow-x: hidden;
+  }
+  
+  .content-overflow {
```
lperry25, line 845: where the important tags required here and on the mjx-container svg?

PR #3793, opened 2025-04-24
add documentation for "filter" keyword

arnej27959:
@bjorncs please review
@kkraune please review also
Note: I've added "limit 0" a couple of places to get only the grouping results.  Also, the "Select parameter language grammar" title was confusing when our documentation mostly doesn't show grouping sent as "select" parameter.


bjorncs (approved):

en/grouping.html
```diff
@@ -1447,6 +1447,15 @@ <h2 id="nested-groups">Nested Groups</h2>
 <h2 id="structured-grouping">Structured grouping</h2>
 <p>
 Structured grouping is nested grouping over an array of structs or maps.
+In this case, each array element is treated as a sub-document and may
+be grouped separately. See the reference for grouping on
```
kkraune, line 1451: I added https://github.com/vespa-engine/documentation/pull/3796 also today, so we might look at linking all this

kkraune (approved):
Great, thanks!

PR #3809, opened 2025-04-30
Add break word and width to table styles

vadimkatsel:
- new styles which prevents too little first cols in tables

### DIFF
--- a/_includes/js.html
+++ b/_includes/js.html
@@ -7,27 +7,3 @@
 
 <!-- Plausible analytics tracking - third-party tracking, no cookies, just for our internal tracking use -->
 <script defer data-domain="vespa.ai,docs.vespa.ai" src="https://plausible.io/js/script.js"></script>
-
-<!-- Script to enabla width scroll if they're getting too big -->>
-<script>
-    window.addEventListener('DOMContentLoaded', () => {
-    const tables = document.querySelectorAll('table');
-  
-    tables.forEach(table => {
-      const parent = table.parentElement;
-      const tableWidth = table.offsetWidth;
-      const parentWidth = parent.offsetWidth;
-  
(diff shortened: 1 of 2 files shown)


css/style.scss
```diff
@@ -881,4 +881,31 @@ mjx-container[display="true"] { // excluding display="inline" and others
...
+    word-break: normal;
+}
+  
+/*Temporary solution of the grid conflict */
+  .lg-col-3-12 {
+    width: 25% !important;
+}
+.lg-col-6-12 {
+    width: 50% !important;
+}
```
vadimkatsel, line 911: I understand how unsatisfied we could be with this solution, but it seems the only one way to solve the problem quickly.

lperry25: yeah i think we should look a bit more at this solution, but I understand it's a quick fix right now to improve the columns issue, but let's not keep this in for long. This could add a lot of frustration in the future I think

_includes/js.html
```diff
@@ -7,3 +7,27 @@
 
 <!-- Plausible analytics tracking - third-party tracking, no cookies, just for our internal tracking use -->
 <script defer data-domain="vespa.ai,docs.vespa.ai" src="https://plausible.io/js/script.js"></script>
+
+<!-- Script to enabla width scroll if they're getting too big -->>
+<script>
```
lperry25: This block of code and the container class you added  can be replaced by using a more specific css selector to only make sure tables that are not already in tables are scrollable. I will make a commit and merge it so we can have this fix live

```/* This excludes tables that are not inside other tables */
table:not(table table) {
    display: block;
    overflow-x: auto;
    white-space: nowrap;
}

lperry25 (approved):

PR #3818, opened 2025-05-02
Remove stale/invalid info about process pid files

hmusum:
Some other minor improvements

There are pid files only for sentinel, config proxy and config server, so doc was wrong. I think it's enough to know how to start and stop services and config server.

### DIFF
--- a/en/operations-selfhosted/admin-procedures.html
+++ b/en/operations-selfhosted/admin-procedures.html
@@ -72,7 +72,7 @@ and is the recommended way to stop Vespa on a node." %}
 
 <h2 id="status-pages">Status pages</h2>
 <p>
-  All Vespa services have status pages, for showing healt, Vespa version, config and metrics.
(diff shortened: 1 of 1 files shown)


en/operations-selfhosted/admin-procedures.html
```diff
@@ -70,23 +70,9 @@ <h2 id="system-status">System status</h2>
...
-</p><p>
-  Learn how to <a href="#vespa-start-stop-restart">start / stop</a> nodes.
-</p>
-
-
-
 <h2 id="status-pages">Status pages</h2>
 <p>
-  Vespa service instances have status pages for debugging and testing.
+  All Vespa services have status pages, for showing healt, Vespa version, config and metrics.
```
kkraune, line 75:
```suggestion
  All Vespa services have status pages, for showing health, Vespa version, config, and metrics.
```

kkraune (approved):
👍

PR #3822, opened 2025-05-02
Add 'linguistics' query.type

bratseth:


en/reference/simple-query-language-reference.html
```diff
@@ -42,6 +42,15 @@
           languages is not supported</li>
     </ul></td>
   </tr>
+  <tr>
+    <th>linguistics</th>
+    <td>Pass the entire string as-is to the linguistics tokenizer and collect all tokens returned into a
+      weakAnd item. This is suitable for applications that want full control over the linguistics processing,
+      typically through a configured linguistics component.
+      When this is used all aspects of linguistics prcoessing including stemming and normalization is controlled
```
arnej27959, line 50: prcoessing -> processing

en/reference/simple-query-language-reference.html
```diff
@@ -42,6 +42,15 @@
...
+      by the linguistics component, and such settings on schema fields are disregarded.
```
arnej27959, line 51: I didn't see any setNormalizable(false), so in a way that seems wrong, but normalization is delegated to linguistics anyway:
`= linguistics.getTransformer().accentDrop(word.getWord(), language);`
the setting on the schema field will still control whether that happens or not, I think.

arnej27959 (approved):
minor spelling mistake; the normalization part I'm unsure about.

en/reference/simple-query-language-reference.html
bratseth: https://github.com/vespa-engine/vespa/pull/33955

PR #3827, opened 2025-05-05
Headings from commented sections in TOC fix

vadimkatsel:
Solution is regex which excludes `<!--`,` -->` and anything between them from content to parse


kkraune (approved):
lets try this

PR #3834, opened 2025-05-05
Excluding commented section's headings from TOC  

vadimkatsel:
Solution is conditions in parser which exclude `<!--`,` -->` and anything between them from content to parse.

**Before:**
image
**After:**
image


kkraune (approved):
lets try this

PR #3840, opened 2025-05-06
Tables Fix: Convert NativeRank reference page tables to lists

vadimkatsel:
Each table now has its own script so it can be independently disabled or re-enabled at any time.


kkraune (approved):
👍

PR #3847, opened 2025-05-07
Podman: Set memory, use crane to get latest version

bjormel:


andreer:
Thanks @bjormel, but after testing now I've discovered that the --memory flag only _limits_ memory - if you say --memory 4G doesn't warn you if there is only 2GB available. So I don't think we should add that, but the crane thing is a very nice idea and we should merge that 👍

andreer (approved):

kkraune:
not sure about crane:
```
zsh: command not found: crane
```
having more dependencies will make this harder to try out ...

PR #3849, opened 2025-05-07
move from root

kkraune:


andreer (approved):

andreer:
Good that we have the redirect, as this url is included in the error messages in previous image versions 👍 
We should probably also update to the new url there: https://github.com/vespa-engine/vespa/blob/master/client/go/internal/osutil/run_cmd.go#L41

PR #3876, opened 2025-05-19
Add main logic and UI of grouping interface

vadimkatsel:
**WARNING:** this feature is fresh, so would be glad if you'll test it out as much as It could be possible before approving. I've tried it with all use cases on the page and haven't noticed anything wierd, but if you will, pleas let me know.

### In this PR:
- new logic to grouping query executor were added.
- new UI were implemented.

### DIFF
--- a/en/grouping.html
+++ b/en/grouping.html
@@ -72,7 +72,14 @@ function showResult() {
         }
         table.classList.add('table');
       } else {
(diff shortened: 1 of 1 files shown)


kkraune (approved):

vadimkatsel:
**NOTICE** There are several cells in Range grouping and Aggregating over all documents, which are getting 403 responses.

PR #3879, opened 2025-05-22
remove documentation about user-defined annotation types

arnej27959:
@kkraune please review
@bratseth FYI


kkraune (approved):
Thanks! great cleanup!

PR #3892, opened 2025-05-30
added example endpoint comment

YashasviMantha:
I should have probably proposed this on slack first but I took the liberty of adding a comment on the external LLMs config page to indicate expected url needs the `v1` path. This is something I myself got confused with while referring to this page configuring ollama and vllm services as a client.

Not sure if this might be confusing for other usecases so pleases feel free to reject the request.  


glebashnik (approved):
Looks good!
Thanks @YashasviMantha.

PR #3897, opened 2025-06-02
document select-elements-by

havardpe:
@arnej27959 please review
@toregge @kkraune FYI


arnej27959 (approved):
this works for weighted set and maps only with elementwise features, not tensors with labels - should maybe be explained here.

PR #3904, opened 2025-06-04
Add use of maintenance mode to admin procedure documentation

vekterli:
@kkraune please review

Not knowing the subtle and fine art of using `maintenance` mode when temporarily taking content nodes down is a frequent occurrence amongst users, which isn't too strange since you really have to know where to look in order to learn about it. To make things a bit better, mention the use of `vespa-set-node-state` explicitly under the ops procedures for stopping and starting a Vespa node.

### DIFF
--- a/en/operations-selfhosted/admin-procedures.html
+++ b/en/operations-selfhosted/admin-procedures.html
@@ -49,8 +49,32 @@ and is the recommended way to stop Vespa on a node." %}
   See <a href="multinode-systems.html#aws-ec2">multinode</a> for <em>systemd</em> /<em>systemctl</em> examples.
   <a href="/en/operations-selfhosted/docker-containers.html">Docker containers</a> has relevant start/stop information, too.
 </p>
-
-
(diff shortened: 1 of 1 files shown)


en/operations-selfhosted/admin-procedures.html
```diff
@@ -49,8 +49,32 @@ <h2 id="vespa-start-stop-restart">Vespa start / stop / restart</h2>
   See <a href="multinode-systems.html#aws-ec2">multinode</a> for <em>systemd</em> /<em>systemctl</em> examples.
   <a href="/en/operations-selfhosted/docker-containers.html">Docker containers</a> has relevant start/stop information, too.
 </p>
-
-
+<h3>Content node maintenance mode</h3>
```
kkraune, line 52:
```suggestion
<h3 id="content-node-maintenance-mode">Content node maintenance mode</h3>
```

kkraune (approved):
👍

PR #3905, opened 2025-06-04
Remove select-mode functionality

vadimkatsel:
Commented mode selection of the docs.


lperry25 (approved):

lperry25:
@kkraune fyi I thought we should remove this since we are not using it and it is adding to the issues with our flashing side bar. Let me know if you don't agree with removing it

vadimkatsel:
@kkraune **fyi:** it also adds the problems to the printed version of the pages it’s presented on, so that it ends up covering the title.

kkraune (approved):

PR #3912, opened 2025-06-05
Remove sidebar from printed version

vadimkatsel:
So that's how printed version of the first page looks like now:

image

Is there any edits wanted, please, let me know.

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -196,6 +196,12 @@ button.section-searchbar {
         height: 473px;
     }
 }
+/* Printed version styling */
+@media print {
+    #sidebar {
+        display: none;
+    }
+}
 
 /* Guides */
 .section-guide--item {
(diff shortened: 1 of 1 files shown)


css/style.scss
```diff
@@ -466,7 +472,10 @@ pre > code {
     margin: 2px 16px;
     
 }
-
+a.tab-link {
```
lperry25, line 475: is this change related to the printed version?

ean:
Is it possible to remove the hamburger-icon as well?

css/style.scss
vadimkatsel: No, thanks for noticing it. Removed. : )

vadimkatsel:
> Maybe some kind of styling would make it more clear that this is a Vespa Cloud-feature?
I can try to play with styling, at the very least as @lperry25 said we can just remove it from printed version.

ean:
Removing it sounds good to me!

lperry25 (approved):

PR #3930, opened 2025-06-12
Add documentation for private model hub embedders

evgiz:
Documents the usage of `secret-ref` for `model` type configurations (i.e support for private models in Huggingface).

### DIFF
--- a/en/embedding.html
+++ b/en/embedding.html
@@ -171,7 +171,7 @@ to produce tokens which are then input to a supplied transformer model in <a hre
 <code>url</code> to supply them from a remote server, or
 <code>model-id</code> to use a
 <a href="https://cloud.vespa.ai/en/model-hub#hugging-face-embedder">model supplied by Vespa Cloud</a>.
-You can also use a model hosted in private Huggingface Model Hub by configuring adding your Huggingface API token
+You can also use a model hosted in private Huggingface Model Hub by adding your Huggingface API token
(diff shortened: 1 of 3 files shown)


en/embedding.html
```diff
@@ -170,8 +170,11 @@ <h3 id="huggingface-embedder">Huggingface Embedder</h3>
 <p>Use <code>path</code> to supply the model files from the application package,
 <code>url</code> to supply them from a remote server, or
 <code>model-id</code> to use a
-<a href="https://cloud.vespa.ai/en/model-hub#hugging-face-embedder">model supplied by Vespa Cloud</a>,
-see <a href="reference/embedding-reference.html#model-config-reference">model config reference</a>.
+<a href="https://cloud.vespa.ai/en/model-hub#hugging-face-embedder">model supplied by Vespa Cloud</a>.
+You can also use a model hosted in private Huggingface Model Hub by configuring a secret
```
onurkaracali, line 174: 1. I believe that it's not any secret, we could state that it's the auth token defined in hugging face. Even we can add a link to hugging face docs to how to create a token. We can guide the reader, and after creating the token, add save it in the cloud secret store and define the secret id in service xml.

2. We could also add a snippet with a private model and secret to show to the users.

onurkaracali: I see the snippet in the embedding-reference..

evgiz: Yeah 😄

PR #3931, opened 2025-06-13
Fix doc of default config values for flush strategy

hmusum:
See default values in https://github.com/vespa-engine/vespa/blob/97d25065dc1b5ed3669c7b6fcc6488e67ad969e3/configdefinitions/src/vespa/proton.def#L57-L68


hmusum:
@toregge please review

PR #3970, opened 2025-07-02
Add preliminary Vespa 9 release notes

hmusum:
Draft, there are no links to this from anywhere yet.

### DIFF
--- a/en/vespa9-release-notes.html
+++ b/en/vespa9-release-notes.html
@@ -1,0 +1,350 @@
+---
+# Copyright Vespa.ai. All rights reserved.
+title: "Vespa 9 Release Notes"
+---
+
+<p>
+  This document lists the changes between Vespa major versions 8 and 9.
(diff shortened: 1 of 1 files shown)


en/vespa9-release-notes.html
```diff
@@ -0,0 +1,350 @@
...
+
+
+
+<h2 id="container-runtime">Container Runtime Environment</h2>
+
+
+<h3 id="jdk-version">JDK version</h3>
+<p>
+  Vespa 9 upgrades the JDK version from 17 to 25 (?). To ensure full compatibility, all container components
+  should be rebuilt with JDK 25 before being deployed on Vespa 9.
```
bjorncs, line 221: Java artifacts built against older JDK versions will still be compatible with Vespa 9 (JDK 25). The opposite will not be possible - Vespa 8 (JDK 21) is not compatible with newer JVM byte code. It's possible though to use the   `--release` option for `javac` to target an older JDK version.

bjorncs (approved):

PR #3990, opened 2025-07-17
Updates news-1-getting-started tutorial so User has their Vespa CLI targeting localhost

michaelokarimia:
Adds an explicit instruction to ensure the user's Vespa CLI is targeting localhost for all subsequent commands in the rest of this tutorial. This will prevent inexperienced users from being unable to get the `vespa status deploy --wait 300` step to pass due to their Vespa CLI being configured for cloud access.


bratseth (approved):
Thanks!

PR #4003, opened 2025-07-28
Rank Feature Reference: add scroll to an anchor after the list has been generated

vadimkatsel:
- add scrollToAnchor function after the table was converted.


kkraune (approved):
👍

PR #4017, opened 2025-07-30
Alternative to global documents

kkraune:


bratseth (approved):
Better, but I think it would be even better to just remove the previous section than adding a new one below which says to not do what it suggests.

kkraune:
ideally we would provide a better guide without parent docs, for model lifecycle, but that is much more work

PR #4022, opened 2025-08-04
Remove small gap from the navbar on mobile

vadimkatsel:
### Description: 
During testing freshly added navbar extension a new problem was noticed:
> image
>
> you can see the content at the very top between the menu and top bar.
### The problem: 
image

 container had its own styles as well as logo did, so both were breaking the grid.

### Solution: 
All the styling was moved to the container.

### Result:

image


kkraune (approved):
👍

PR #4072, opened 2025-08-22
Use correct Osano script, and remove widget

ean:


ean:
Need to fix denali theme color overrides for the Osano views, and add "Cookie Preference" link

frodelu (approved):

PR #4078, opened 2025-08-25
Fix/html tag id must be lowercased

dainiusjocas:

In the `search.vespa.ai` links are all lowercased, while in actual HTM tag ID is camel-cased. Lowercase the tag so that links work in search.


kkraune (approved):
👍

PR #4088, opened 2025-08-29
expand on sync-transactionlog=false

radu-gheorghe:
Expands a bit on the implications and on the tunables.

This might be needed for people considering sync-transactionlog=false. This was something I was doing quite often with Elasticsearch, when disk IO was problematic under heavy writes. Elasticsearch/OpenSearch allow you to configure the async flush behavior, but I don't think this is needed with Vespa, because the kernel exposes even more knobs.


kkraune (approved):
👍

PR #4133, opened 2025-09-26
Add information on state/v1/initialization endpoint

boeker:
This adds information on the state/v1/initialization endpoint to the Proton page and a small remark on this to the state/v1 page.


toregge:
@kkraune : FYI

PR #4146, opened 2025-10-02
Add more info about using index and attribute for a field

hmusum:
Maybe this is a start? @kkraune please review


en/text-matching.html
```diff
@@ -45,6 +43,24 @@ <h2 id="index-and-attribute">Index and attribute</h2>
...
+  If both index and attribute are configured for string-type fields, Vespa will search and match against the
+  index with default match <code>text</code>.   When the field is both index and attribute, the index
+  aspect is used for matching. (The attribute could still be useful in general, for grouping and sorting.)
+  To get multiple match modes on a single source field you could define a synthetic field:
+</p>
+<pre>
+field album type string {
+    indexing: summary | index
+}
+field album_as_attribute type string {
```
kkraune, line 56: a common error is having the synthetic fields inside the document block, so the example should include this, too - the current snippet will not work

kkraune:
👍

PR #4638, opened 2026-04-13
Document <max-document-size> element under <document-api>

hmusum:
## Summary
- Document the new optional `<max-document-size>` child element of `<document-api>` in `services.xml`, added in vespa-engine/vespa#36391.
- Describes the request-size limit (1–2048 MiB, default 128 MiB), MiB-preferred unit syntax, and the deploy-time warning when the value exceeds any content cluster's `max-document-size`.
- Updates the `<document-api>` TOC and combined example.


glebashnik:
@claude

glebashnik (approved):
Looks good!

PR #4646, opened 2026-04-15
Update rank feature doc with featureName(fieldName) parameter syntax

arnej27959:
Vespa now supports an alternative naming convention for rank feature configuration parameters: instead of featureName.paramName.fieldName, you can use featureName(fieldName).paramName. This is more consistent with how rank features are referenced elsewhere (e.g. in ranking expressions).

Update the nativerank and rank-feature-configuration reference docs to use the new syntax throughout. Also add cross-reference rows for nativeRank, nativeFieldMatch, and nativeProximity in the configuration table, document elementwise bm25 fallback behavior, and modernize HTML entities to use <em> consistently.


bratseth (approved):
Great!

kkraune (commented):
👍

PR #4648, opened 2026-04-16
Add restore procedure documentation

evgiz:


en/operations/data-management.html
```diff
@@ -26,13 +26,38 @@ <h2 id="backup">Automated Backups</h2>
...
+<p>
+  Restore requires a deployed target cluster with:
+</p>
+<ul>
+  <li>The same number of content nodes as the backup.</li>
+  <li>At least equivalent disk capacity per node as at the time of the backup.</li>
+</ul>
+<p>
+  The cluster must be active at the time of the restore.
+  Note that <code>cluster</code> granularity backups restore with zero content redistribution,
```
olaaun: There is zero redistribution in the ideal scenario, but this is only true if the set of distribution keys is equal during backup and restoration time.

This is changing soon, but that change will also remove the redistribution during grouped restoration.

evgiz: Good point, I'll update

olaaun (approved):

kkraune (commented):
👍

PR #4664, opened 2026-04-22
clarify

kkraune:


hmusum (approved):

bratseth:
Imho duplicating info on how the relevance is calculated here is unnecessary (and in general the final value can also be calculated in some other way, e.g. if there are external third-phase rankers).

kkraune:
true. there was a user asking, from the result json page, where does this value come from. so maybe better to put a small section in https://docs.vespa.ai/en/ranking/ranking-intro.html and link to there from result json doc. It was not clear to him. and we can put the Nan stuff there too

bratseth:
I think the user question turned out to be more "why do I get this error", i.e. the NaN thing. Yes, we could point to ranking-intro with the word "relevance". That doc should be rewritten a lot btw ...

PR #4668, opened 2026-04-24
Add more details for -Inf

kkraune:


en/learn/faq.md
```diff
@@ -81,13 +81,39 @@ The additional filters are applied as a post filtering
...
+function finite_or_sentinel(x) {
+      expression: if (isNan(x - x), -1e9, x)
+}
+    ```
+- Use CBOR instead of JSON - using a binary format can represent NaNs and Infinities without issues,
+  and it can also be faster/more efficient
 
-- The [ranking](../basics/ranking.html) expression used a feature which became `NaN` (Not a Number). For example, `log(0)` would produce
--Infinity. One can use [isNan](../reference/ranking/ranking-expressions.html#isnan-x) to guard against this.
-- Surfacing low scoring hits using [grouping](../querying/grouping.html), that is, rendering low ranking hits with `each(output(summary()))` that are outside what Vespa computed and caches on a heap. This is controlled by the [total-keep-rank-count](../reference/schemas/schemas.html#total-keep-rank-count) parameter.
```
andreer, line 90: I don't understand how this is related - is this supposed to be a separate FAQ entry that is being removed?

andreer (approved):

en/learn/faq.md
kkraune: I was puzzled, too ... let's find out in next pr

PR #4675, opened 2026-04-28
docs: document <prepend> config for OpenAI embedder

bjorncs:
## Summary

- Documents the new `<prepend>` element for the OpenAI embedder in `en/reference/rag/embedding.html` (reference table row with example).
- Adds a cross-reference from the "Adding a fixed string to a query text" section in `en/rag/embedding.html` so users searching for prepend find that it also works for the OpenAI embedder.
- Pairs with the implementation PR: vespa-engine/vespa#36580.

### DIFF
--- a/en/rag/embedding.html
+++ b/en/rag/embedding.html
@@ -733,6 +733,11 @@ emit the same predefined metrics, although emitting custom metrics is perfectly 
   sample application.
 </p>
 <p>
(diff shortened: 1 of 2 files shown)


thomasht86 (approved):

glebashnik (approved):

kkraune (commented):
👍

PR #4678, opened 2026-04-28
Update docs with default value for max-document-size

hmusum:
Also point FAQ to the new config that can be set for document api


glebashnik (approved):
Looks good

PR #4685, opened 2026-05-05
Add firstPhaseMax

johansolbakken:
Adds an entry for `firstPhaseMax`.

### DIFF
--- a/en/reference/ranking/rank-features.html
+++ b/en/reference/ranking/rank-features.html
@@ -1286,6 +1286,15 @@ tensor(id{}):{ {id:100}:10.5, {id:200}:20.75, {id:300}:30.25 }
     </p>
   </td></tr>
 
+<tr><td>firstPhaseMax</td>
+  <td>-infinity</td>
+  <td>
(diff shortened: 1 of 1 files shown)


en/reference/ranking/rank-features.html
```diff
@@ -1286,6 +1286,15 @@ <h3 id="rank-scores">Rank scores</h3></td></tr>
     </p>
   </td></tr>
 
+<tr><td>firstPhaseMax</td>
+  <td>-infinity</td>
+  <td>
+    <p id="firstPhaseMax">
+      The maximum first-phase rank score among all hits matched on the content node.
+      Calculated locally on each content node, so may yield different results ondifferent nodes.
```
hmusum:
```suggestion
      Calculated locally on each content node, so may yield different results on different nodes.
```

hmusum (approved):
👍  (one typo, fix suggested)

en/reference/ranking/rank-features.html
johansolbakken: Thanks!

PR #4693, opened 2026-05-07
Easier navigation to the variants

kkraune:
... i always go to the services.xml reference for this, not to find it ...

I consider moving this from deployment-variants.html into the services.xml reference, and the rest into some query profile reference - it feels odd to have it as a separate doc, with two different things in it - what do you think @hmusum  ?


hmusum (approved):
Yeah, I agree, I have the same issue.

 I think this is good as long as we think it's OK to have cloud-specific stuff in this doc

PR #4696, opened 2026-05-08
Use ruby 3.4.9 - fix linkchecker

kkraune:
```
bundle install 
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Could not find compatible versions

Because github-pages >= 232 depends on jekyll-commonmark-ghpages = 0.5.1
  and jekyll-commonmark-ghpages >= 0.2.0 depends on jekyll-commonmark ~> 1.4.0,
  github-pages >= 232 requires jekyll-commonmark ~> 1.4.0.
And because jekyll-commonmark >= 1.4.0 depends on commonmarker ~> 0.22
  and commonmarker >= 0.22.0, < 1.0.0.pre depends on Ruby >= 2.6, < 4.0,
  github-pages >= 232 requires Ruby >= 2.6, < 4.0.
So, because Gemfile depends on github-pages >= 232
  and current Ruby version is = 4.0.2,
  version solving has failed.
```

### DIFF
--- a/.github/workflows/build-pages.yml
+++ b/.github/workflows/build-pages.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: "3.4.8"
+          ruby-version: "3.4.9"
           bundler-cache: true
           cache-version: 0
       - name: Build with Jekyll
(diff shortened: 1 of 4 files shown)


kkraune:
I did not fix it at all... I'll leave this for some time until new doc serving is in place and we can then drop GH pages and problem is solved

PR #4704, opened 2026-05-13
Select syntax: doc type selection example

radu-gheorghe:
And, in general, that you could use other parameters.


bratseth:
Ok I guess but in general we want the reference doc to only provide the raw information and put examples and other explanation in guides.

radu-gheorghe:
I don't think we have a Select guide, though :(

bratseth:
True ... so feel free to merge.

PR #4708, opened 2026-05-15
Standardize spelling of "document ID"

boeker:
Use spelling variant "document ID". If referring to a heading, follow capitalization of the heading. This applies only to text, of course. The internal field is called `documentid`. Sometimes `id` is used. These should stay the way they are.

@toregge let me know if I was to eager somewhere 🙂 or not eager enough


toregge (approved):
@kkraune : FYI

PR #4721, opened 2026-05-21
docs: add MCP server sections to llm-help

edvardwd:

Since some MCP capabilities were created some time ago, I feel like this is a more accurate description. Also think these headers make more sense as it would be natural to divide into public MCP server (like a doc search e.g.) and "private" where you can interact with your own app.


glebashnik (approved):
Looks good.

PR #4730, opened 2026-05-27
Be more explicit about documentid setting requiring restart

boeker:
Makes it clear that the `documentid` setting will not trigger a restart automatically on Vespa Cloud.


hmusum (approved):
Nice, thanks!

PR #4736, opened 2026-05-28
Kkraune/instance types

kkraune:


en/performance/node-resources.md
```diff
@@ -111,9 +111,32 @@ with **zero manual interventions**.
...
+Observe variations in both architecture and memory capacity.
+CPU architectures have different load profiles,
+but not setting the architecture increases likelihood of successfully provisioning the node during shortages.
+
+Also, a 256G RAM node has more memory than a 252G RAM node.
+The same argument applies, so Vespa Cloud allows a 2% variation in order to select from more options:
+
+    <resources vcpu="32" memory="252Gb">
+
+This variation in one-way - a node can have 2% _more_ resources, not less -
```
bratseth: No it's 2-way.

en/performance/node-resources.md
```diff
@@ -111,9 +111,32 @@ with **zero manual interventions**.
...
-[AWS](https://cloud.vespa.ai/en/reference/aws-flavors.html),
-[GCP](https://cloud.vespa.ai/en/reference/gcp-flavors.html), and
-[Azure](https://cloud.vespa.ai/en/reference/azure-flavors.html) instance types.
+[AWS](instance-types/aws-instance-types.html),
+[GCP](instance-types/gcp-instance-types.html), and
+[Azure](instance-types/azure-instance-types.html) instance types.
+
+
+### Instance type equivalence
+Instance types come with variations -
```
bratseth: I suggest "Some instance types are so similar that it does not matter which one you get. For example ..."

bratseth:
... one more thought: Maybe we don't need to be this verbose. Can't we just say "Node must match resources to within a 2% difference to be eligible." and that's it?

en/performance/node-resources.md
kkraune: ok, that makes it easier, will simplify

PR #4770, opened 2026-06-22
document averageFieldLength

radu-gheorghe:
As introduced by https://github.com/vespa-engine/vespa/pull/36599


bratseth (approved):
Great!

PR #4778, opened 2026-06-25
Document zones with more than one availability zone

hakonhall:
Documents...
* deployment.xml syntax to specify availability zones
* availability zones in our zones page. Also adds 2 multi-AZ zones (the one in prod.aws-eu-west-1 is expected to come online before the end of the month)
* Adds a separate page on availability zones, and the semantics of multi-AZ deployments.

The new availability zone page:
image


kkraune (approved):
💯

PR #4792, opened 2026-07-06
Document JSON grouping labels

radu-gheorghe:
Also documented the difference between group list label and aggregator label.

Corresponds to the change in https://github.com/vespa-engine/vespa/pull/36887


johansolbakken (approved):
👍🏻

PR #4798, opened 2026-07-13
Add maintenance-only block-change example

lperry25:
Adds an example to the `block-change` reference showing how to configure a maintenance-only block window.

A customer wrote `<block-change maintenance="true"/>` expecting to block only maintenance, but since `revision` and `version` default to `true` when omitted, that line also blocked application deployments and platform upgrades during the window. The attribute table documents the defaults, but there is no example combining them, so the mistake is easy to make. The new example spells out that a maintenance-only window needs `revision="false" version="false"`, and shows a config that keeps the maintenance window open well beyond the required 17 hours per week.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kkraune (commented):
Thanks!

PR #4821, opened 2026-08-10
Add guide for batch update

kkraune:
add example for iterating over an update command

FYI @BrageHK

### DIFF
--- a/en/writing/document-v1-api-guide.html
+++ b/en/writing/document-v1-api-guide.html
@@ -523,12 +523,14 @@ $ curl -X DELETE \
 
 </ul>
 
+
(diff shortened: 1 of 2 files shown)


thomasht86:
Example uses only plain curl with bearer token - maybe we should have the example use vespa curl that works with mTLS as well?

kkraune:
> Example uses only plain curl with bearer token - maybe we should have the example use vespa curl that works with mTLS as well?

good idea, added mtls alternative in a comment

thomasht86 (approved):
Great!

BrageHK (approved):

PR #4831, opened 2026-08-18
Shorten and remove some 'advanced' details

bratseth:
I want to keep the documents in basics/ short and simple.

Does this cover what you needed to add? We may need a content-clusters doc under content/.


kkraune:
I added these based on user questions earlier this week, and the LLMs were not able to provide good answers. I am happy to move the multischema elsewhere, though, the models will pick it up regardless. should we have an 'advanced' schema guide?

bratseth:
What I removed is around how schemas map to content clusters and how data is routed to them, so a document on content clusters under content/? I think much of the text in the content-node document could be moved there as well.

kkraune:
makes sense - let me create that instead and we can merge this one for now

PR #4832, opened 2026-08-19
restore the homepage that was added in july and not applied

lperry25:


lperry25:
Now staging is using the versions from the root, so this can be closed

PR #4839, opened 2026-08-21
Add missing fieldTermMatch outputs

bratseth:
@arnej27959 to make WordAlternative exactness generally useful I think we need to use it in various other rank features and create a few new ones. If you agree i can create a ticket on that.

@kkraune @radu-gheorghe fyi


arnej27959:
agreed - my original suggestion (10 years ago?) was to multiply with the exactness in bm25 and nativerank, but it was impossible to get permission to change anything like that...

PR #4842, opened 2026-08-24
Document the YQL labeled() query operator

arnej27959:
Vespa 8.743.16 adds labeled(subQuery, "label", score), which wraps an arbitrary sub-query in a labeled item carrying a raw score. Until now labels could only be attached with the {label: "..."} annotation, and there was no way to give a sub-query an explicit score for ranking.

Add a reference entry for the operator in the YQL query operator table, next to rank(), explaining that the score is exposed to ranking expressions through itemRawScore(label). The example shows the intended use: retrieve on the main query with rank(), then let labeled() clauses contribute boosts without affecting matching, combined them in a rank profile expression.

### DIFF
--- a/en/reference/querying/yql.html
+++ b/en/reference/querying/yql.html
@@ -771,6 +771,30 @@ where rank(nearestNeighbor(field, queryVector), a contains "A", b contains "B", 
   </tr>
 
   <tr>
+    <th>labeled</th>
+    <td>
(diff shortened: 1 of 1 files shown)


kkraune (approved):
great, and the example is useful!

bratseth (approved):
👌

PR #4844, opened 2026-08-25
Relate labeled() to equivalent operators in other search engines

arnej27959:
Add a short note to the labeled() reference documentation pointing out that it plays the same role as Lucene's ConstantScoreQuery, reached via function_score in Elasticsearch/OpenSearch and via the ^= syntax in Solr.

Why: users migrating from those engines look for a way to attach a fixed score contribution to a sub-query without letting it affect matching. Naming the operator they already know makes labeled() discoverable from that starting point, instead of leaving them to infer the mapping from the Vespa-specific description of labels and itemRawScore.


bratseth (approved):
👍

kkraune (commented):
👍

PR #4845, opened 2026-08-26
docs: production tests

BrageHK:

Wait for merge until feature has converged and is properly working on test application.

### DIFF
--- a/en/reference/applications/testing-production.html
+++ b/en/reference/applications/testing-production.html
@@ -1,0 +1,271 @@
+---
+# Copyright Vespa.ai. All rights reserved.
+title: Production tests
+---
+
+<p>
+  This is the Vespa Testing reference for <a href="../../applications/testing.html">production tests</a>.
+</p>
+<p>
+  Production tests verify domain specific metrics for a production deployment, such as query latency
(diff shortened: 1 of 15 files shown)


en/reference/applications/testing-production.html
```diff
@@ -0,0 +1,271 @@
+---
+# Copyright Vespa.ai. All rights reserved.
+title: Production tests
+---
+
+<p>
+  This is the Vespa Testing reference for <a href="../../applications/testing.html">production tests</a>.
+</p>
```
kkraune, line 9: I think users will put prod tests in any prod zone before moving on to the next, too. The canary case is also true, but we should rewrite to generally validate the zone before going to next

en/reference/applications/testing-production.html
```diff
@@ -0,0 +1,271 @@
+---
+# Copyright Vespa.ai. All rights reserved.
+title: Production tests
```
kkraune, line 3: we should have the cloud annotation here

kkraune (approved):
nice! I did not fine-read all the details, we can iterate on this

BrageHK:
I have now updated the documentation to comply with the most recent version of the prod tests. @olaaun @kkraune

kkraune (approved):
👍

PR #4855, opened 2026-09-02
Add guidance on aggregating over large result sets

thomasht86:
Grouping over a large matched set, such as category counts over the full corpus, is a common thing users want to do, causing issues if done frequently.

Adds a section to the grouping guide describing how to reduce the work per query, how to compute approximate counts from a random 1% sample using a synthetic random field, and when to cache aggregates in a Searcher. Link to it from the FAQ and the performance docs.


kkraune (commented):
👍

kkraune (approved):

